PR #3, opened 2013-10-02
catch and ignore OutOfMemoryError in getRotatedBitmap()

kworam:
getRotatedBitmap() can run out of memory if the image is very large:

http://simonmacdonald.blogspot.ca/2012/07/change-to-camera-code-in-phonegap-190.html

If this happens, simply do not rotate the image and return it unmodified.  If you do not catch the OutOfMemoryError, the Android app crashes.


kworam:
Thanks for closing the pull request on my behalf agrieve, I got distracted and forgot all about this for a while.

PR #12, opened 2014-02-12
added allowEdit support for android

riyazMuhammad:
## Changes
1. Allows user to crop selected image from gallery or captured from camera if **allowEdit** is set to **true**
2. Fallback if device does not support **com.android.camera.action.CROP**

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,5 @@
+Apache Cordova
+Copyright 2012 The Apache Software Foundation
+
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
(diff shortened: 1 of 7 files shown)


roysG:
there is still problem with cropping in android, when  choose to capture image.
I can׳t crop, why ?

fguerin:
Same as @roysG, I'm not able to edit the captured image.

jsappme:
Same as @roysG, I'm not able to edit the captured image. I have a Nexus 4.

jcesarmobile:
Read the quirks

Android Quirks
Android also uses the Crop Activity for allowEdit, even though crop should work and actually pass the cropped image back to Cordova, the only one that works consistently is the one bundled with the Google Plus Photos application. Other crops may not work.

So, you have any app that can crop like google plus photos? in case you have it, does it open?

PR #24, opened 2014-04-30
[iOS] Returns a specific error message when app has no access to Photos

m-revetria:
If user cancel the UIImagePickerController and the app has no authorization to access to the Photos, an error with the message "has no access to assets" is returned as the plugin result.

### DIFF
--- a/src/ios/CDVCamera.m
+++ b/src/ios/CDVCamera.m
@@ -394,7 +394,7 @@ static NSSet* org_apache_cordova_validArrowDirections;
     // popoverControllerDidDismissPopover:(id)popoverController is called if popover is cancelled
 
     CDVPluginResult* result;
-    if ([ALAssetsLibrary authorizationStatus] == ALAuthorizationStatusAuthorized) {
+    if (picker.sourceType == UIImagePickerControllerSourceTypeCamera || [ALAssetsLibrary authorizationStatus] == ALAuthorizationStatusAuthorized) {
         result = [CDVPluginResult resultWithStatus:CDVCommandStatus_ERROR messageAsString:@"no image selected"];   // error callback expects string ATM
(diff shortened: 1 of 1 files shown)


nleclerc:
Hi, sorry to resurrect an old thread but i'm having trouble with this modification.
When users click cancel in camera interface, we get a `has no access to assets` error instead of `no image selected`.

@m-revetria do you have more information about your use case ?
I fail to see how this test is pertinent here.
If user cancels action then it's just a cancellation, not a auth error.

m-revetria:
Hi, Nicolas.

The use case was show an alert message to the user indicating:

   - If he canceled the selection of an image that he must select an image
   in order to continue.
   - If he canceled *due to the app has no access to the photos
   library/camera*, then show an alert message to let the user knows the
   app wasn't granted to access that resource.


That error could happen in the case where the user previously denied access
to camera roll and later he opens the camera to take a picture but he
cancels. In that case `[ALAssetsLibrary authorizationStatus] !=
ALAuthorizationStatusAuthorized`and the error "has no access to assets" is
returned wrongly.

I have solved this problem in commit:
https://github.com/m-revetria/cordova-plugin-camera/commit/bbdafb08640cafc8e98ed0cbd5d558ad9b6dc150
.

I didn't create a PR with this change yet.

Thanks!



[image: xmartlabs website] <http://xmartlabs.com/>
MIGUEL REVETRIA
iOS Engineer
_________________
<https://www.linkedin.com/in/mrevetria> <https://github.com/m-revetria>

nleclerc:
Thanks for the answer.
Actually i was a bit confused because i came over the PR from another
commit, hence my question.

This new commit seems very nice, i hope it can be merged quickly ! :-)

PR #29, opened 2014-06-04
Fix cb6875

dpesch:
This fix will fix the FATAL EXCEPTION which kills the app process on HTC Android devices (please notice CB6875 https://issues.apache.org/jira/browse/CB-6875).

We've decided to catch the RuntimeException instead of UnsupportedOperationException to catch also the IllegalStateException without doubling the code in the catch block. I'll hope you'll accept this nevertheless.


hearbeathorror:
Hi, i am trying to replicate this exception, can you please let me know how exactly to reproduce this? Does this happen when we have no sdcard present ?

PR #37, opened 2014-07-17
StatusBar Hidden Fix

dariodepaolis:
If I click cancel on imagePicker and I have status bar hidden it re-appear, so I have to hide it also in this case.

### DIFF
--- a/src/ios/CDVCamera.m
+++ b/src/ios/CDVCamera.m
@@ -386,6 +386,8 @@ static NSSet* org_apache_cordova_validArrowDirections;
     }
     
     [self.commandDelegate sendPluginResult:result callbackId:cameraPicker.callbackId];
+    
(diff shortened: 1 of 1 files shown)


shazron:
My apologies, the code has been refactored to be testable, so it doesn't merge cleanly anymore. It should be an easy rebase from your patch (I hope) - and my previous comment needs to be addressed as well.

fretman92:
+1 for this issue, the plugin should detect the presence of the status bar, and hide it when finished if necessary.

janpio:
Hey @dariodepaolis , there seems to be a merge conflict now. Could you maybe fix this? Thanks.

jcesarmobile:
This just hides the statusbar, but not all users will want this.
Also has merge conflict and requested changes have not been done, so I'm going to close it.
I think this should belong to statusbar plugin, probably already handles it.

PR #49, opened 2014-09-30
CB-7667 iOS8: Handle case where camera is not authorized

dpolivy:
In iOS 7+, when the app does not have access to the camera, show a
prompt notifying the user so they're not puzzled by looking at a black
screen.

In iOS 8+, include a link on the dialog to open the Setting app to allow
the user to change their Camera privacy setting.

Fixes https://issues.apache.org/jira/browse/CB-7667

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -146,6 +146,7 @@
          <framework src="AssetsLibrary.framework" />
          <framework src="MobileCoreServices.framework" />
          <framework src="CoreGraphics.framework" />
+         <framework src="AVFoundation.framework" />
          
          <config-file target="*-Info.plist" parent="NSLocationWhenInUseUsageDescription">
              <string></string>
--- a/src/ios/CDVCamera.m
+++ b/src/ios/CDVCamera.m
@@ -24,6 +24,7 @@
 #import <ImageIO/CGImageProperties.h>
 #import <AssetsLibrary/ALAssetRepresentation.h>
 #import <AssetsLibrary/AssetsLibrary.h>
+#import <AVFoundation/AVFoundation.h>
 #import <ImageIO/CGImageSource.h>
 #import <ImageIO/CGImageProperties.h>
 #import <ImageIO/CGImageDestination.h>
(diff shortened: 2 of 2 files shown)


shazron:
My apologies, the code has been refactored to be testable, so it doesn't merge cleanly anymore. It should be an easy rebase from your patch (I hope).

dpolivy:
I cleaned up the merge conflict, and I believe it's now working again. I had a bunch of issues upgrading my plugin (CDVCamera.m seemed to drop off the list of compiled source files?), but got those sorted out and it seems ok. Please do sanity test it on your side as well!

dpolivy:
@shazron I just re-merged master and rebased the changes so it's a clean commit. Any chance of getting this pulled in to the next release?

dpolivy:
@stevengill @shazron Is it possible to get a review on this and some feedback on whether it is OK to merge?

dpolivy:
@shazron Any possibility of a review & merge for the next plugin release?

shazron:
Thanks for the PR Dan -- I'm away, will take a look Monday.

seronal:
I don't know what is the official stance on this, but I don't think it's a good idea to hardcode an error message string (i.e. "Access to the camera has been prohibited; please enable it in the Settings app to continue."). This will be a problem in terms of localization. It should perhaps be passed within CameraOptions?

shazron:
In the end it's not really hardcoded. If you notice, it is wrapped using NSLocalizedString. If you have a localized String resource file, it will load from the key "Access to the camera has been prohibited; please enable it in the Settings app to continue."

seronal:
Yes but not everyone uses native localization support. It's much easier to handle the localization on the javascript side, especially if you are targeting multiple platforms. This approach basically requires an extra plugin to create/move localization resources if you are using CLI based approach for development, not to mention having to create platform specific localized resources.

dpolivy:
As this is iOS specific, I'm not sure it makes sense as a JavaScript parameter. Is there any provision for translating strings that are part of the Cordova project? Perhaps the translated versions can be added to plugin.xml, so they are automatically included in the native resources as necessary when the plugin is installed.

shazron:
I believe localization should never be up to the user of the component; the component itself should handle localization seamlessly.

This is how the Cordova Media Capture plugin does it:
1. https://github.com/apache/cordova-plugin-media-capture/tree/master/src/ios/CDVCapture.bundle
2. https://github.com/apache/cordova-plugin-media-capture/blob/f5514cabe19648a4c952bf05c09e088510895165/src/ios/CDVCapture.m#L39

To add more localizations, just add the appropriate .lproj folder with a Localizable.strings file -- send pull requests for adding each language as volunteers provide the strings.

In cordova-ios 4.0.x (coming in the future), these resource API functions will be available for plugin authors so they can easily localize their plugins: https://github.com/apache/cordova-ios/blob/ddadf8992bb4b34cd728ed9a21e880ad20e551db/CordovaLib/Classes/Public/CDVPlugin%2BResources.m#L24

BKHines:
Question: does this mean if I want the message to be "Hey!  Open Settings to Grant Camera Permission" instead of "Access to the camera has been prohibited; please enable it in the Settings app to continue." I can change it through a string localization?  I have a client who has a desire to change that value to a different phrase and I was close to saying it's not configurable so I wanted to check.

dpolivy:
> Question: does this mean if I want the message to be "Hey! Open Settings to Grant Camera Permission" instead of "Access to the camera has been prohibited; please enable it in the Settings app to continue." I can change it through a string localization? I have a client who has a desire to change that value to a different phrase and I was close to saying it's not configurable so I wanted to check.

Yes, you should be able to override it using a localization file.

BKHines:
Wow.  Interesting.  Thanks for the response; I had never considered doing that before.

PR #54, opened 2014-11-12
cameraDirection option fix in Android

lgrignon:
Uses Android Camera API, which in API level 9, received some feature detection abilities.
A custom activity has been created to control the camera and offer something similar to the the standard Android camera app.


joshstrange:
Sorry if I'm missing something but why was this never merged?

lgrignon:
Hey, I couldn't tellement because nobody answered it but I supposed it was because the cordova plugin system changed between my fork and my PR :)

PR #55, opened 2014-11-21
Update CDVCamera.m

gitMarcus:
If you don't want edit JPEG but reduce quality you can't. With this you can...


shazron:
My apologies, the code has been refactored to be testable, so it doesn't merge cleanly anymore. It should be an easy rebase from your patch (I hope).

PR #56, opened 2014-11-24
Correct image resolution selection

kiwinewt:
Update the code to correctly select the CameraCaptureUIMaxPhotoResolution based on the Windows guidelines.

### DIFF
--- a/src/windows/CameraProxy.js
+++ b/src/windows/CameraProxy.js
@@ -369,16 +369,16 @@ module.exports = {
                 }
 
                 // decide which max pixels should be supported by targetWidth or targetHeight.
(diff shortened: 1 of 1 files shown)


muratsu:
@kiwinewt  thanks for the contribution. The right approach here is to take into account the max num of pixels instead of a certain height or width limit. 

VeryLarge5M = 5 * 1000 * 1000;
 Large3M = 3 * 1000 * 1000;
 MediumXga = 1024 * 768;
 SmallVga = 640 * 480;
 VerySmallQvga = 320 * 240;

If you can modify the code as such, I'll be glad to accept your PR - thanks!

PR #58, opened 2014-12-16
Fix terminology for selecting a video.

coryjthompson:
Selecting a video currently shows 'Videos title' along the top of the picker.

I believe 'Videos' would be more appropriate. 

The reason it was changed was for localization, which unless I am mistaken has been removed.

(Old commit showing it was previously 'Videos' until localization was added. Probably completely irrelevant)
https://github.com/coryjthompson/cordova-ios/commit/8b0f11fd60bf6966bd40d296cd80b96e2cef6806

Thanks.


purplecabbage:
This is actually looking for a localized string for the key "Videos title"
Removing the localization call does not fix the issue.
Since the key does not appear to exist anywhere, I think a good compromise for now would be to change it to use `NSLocalizedString(@"Videos", nil)];` as NSLocalizedString will return the key passed to it if it does not find a localized version.

PR #64, opened 2015-02-17
CB-8498: On windows import from library shall return native URI when requested.

vldmrrr:


Bubkastor:
Image not open this uri object.

src/windows/CameraProxy.js
```diff
@@ -165,7 +165,16 @@ module.exports = {
 
                             var storageFolder = Windows.Storage.ApplicationData.current.localFolder;
                             file.copyAsync(storageFolder, file.name, Windows.Storage.NameCollisionOption.replaceExisting).then(function (storageFile) {
-                                successCallback(URL.createObjectURL(storageFile));
+                                var res;
+                                if (destinationType == Camera.DestinationType.NATIVE_URI) {
+                                    res = storageFile.path
+                                        .replace(/\\/g, '/')
+                                        .replace(':','%3A');
```
Bubkastor: there is no need to replace

vldmrrr: Colon is invalid character in URI. This replacement makes returned URI compatible with native URI suitable for File plugin.

src/windows/CameraProxy.js
```diff
@@ -165,7 +165,16 @@ module.exports = {
 
                             var storageFolder = Windows.Storage.ApplicationData.current.localFolder;
                             file.copyAsync(storageFolder, file.name, Windows.Storage.NameCollisionOption.replaceExisting).then(function (storageFile) {
-                                successCallback(URL.createObjectURL(storageFile));
```
Bubkastor: Maybe just 
return successCallback("ms-appdata:///local/" + storagefile.name);

vldmrrr: Yeah, that should work as well. Just need to test.

vldmrrr: OK, tested, works, amended commit as suggested

PR #67, opened 2015-02-24
CB-8253 Fix potential unreleased resources

akinzie:
There was a place (~line 701) in CameraLauncher.java where there was the
potential for input and output streams to never be closed if an exception
occurs at the wrong time.   There were some other places where an
InputStream was used anonymously, and so would never be closed.

This change introduces try/finally blocks to ensure that the streams will
always end up closed.

Change-Id: I479bceddcd631bfec45c3f5ee7e88ddb04c59073

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -24,6 +24,7 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


vladimir-kotikov:
@akinzie, could you please close this, since it has been merged already in https://github.com/apache/cordova-plugin-camera/pull/90

akinzie:
Sorry, didn't realize this was still open.

PR #68, opened 2015-02-25
CB 8438cordova plugin camera translation

sosahvictor:
CB 8438cordova plugin camera translation: cordova-plugin-camera


sosahvictor:
This will not be merged here

PR #71, opened 2015-03-02
Fix for geolocation/exif data

auser:
Two part pull req:

1. Currently, if the options include geolocation, i.e.:

```xml
<preference name="CameraUsesGeolocation" value="true" />
```

Then the plugin drops into a method that doesn't actually contain any data. This fixes the flow so that data is supplied with geo

2. This pull req also attaches GPS, Exif, and RAW data to the metadata. Previously, no Exif data was actually written back to the image.

### DIFF
--- a/src/ios/CDVCamera.m
+++ b/src/ios/CDVCamera.m
@@ -297,6 +297,7 @@ static NSString* toBase64(NSData* data) {
             } else {
                 data = UIImageJPEGRepresentation(image, [options.quality floatValue] / 100.0f);
             }
+
             if (options.usesGeolocation) {
                 NSDictionary* controllerMetadata = [info objectForKey:@"UIImagePickerControllerMediaMetadata"];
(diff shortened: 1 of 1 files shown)


purplecabbage:
This looks good.  Can you fix the white-space additions?

Also, I don't see you listed on https://people.apache.org/committer-index.html
Can you submit an ICLA? We can pull this with just the whitespace fix, but will need the ICLA for future pulls. details here: http://cordova.apache.org/#contribute

vladimir-kotikov:
@auser, ping

AllenConquest:
@auser @purplecabbage when is this issue going to be fixed? I'm happy to redo the pull request if I can be added the list of committers.

purplecabbage:
@AllenConquest if you have a cla on file then feel free to redo this and send a pr 
@auser do you have a cla on file?

http://cordova.apache.org/contribute/contribute_guidelines.html

PR #73, opened 2015-03-04
Change folder save Windows. And return path instead of blob.

Bubkastor:
Work save standard folder


muratsu:
Most of this functionality is in Camera now. I think you can close this PR.

PR #76, opened 2015-03-06
Put photo path into a file html field

dominosoft:
Hi, is it possible to include automatically the path of the photo to a "file" html field in order to upload that pic to a remote site? if so, how can you do that?


purplecabbage:
No @dominosoft it is not possible.  The file-input can only have it's value set by user interaction which is not available inside a cordova app.

Please close this pull request as it does not make sense for this question to be a pull request ...

PR #81, opened 2015-04-01
CB-8780: Display popover using main thread. Fixes popover slowness

cjpearson:
This moves the UI code in the `takePicture:` method to run on the main thread. This fixes the slowness in displaying the popover. I also changed the remaining `self`s in the block to `weakSelf`.


muratsu:
LGTM.
This also potentially fixes perf issues we have with [CB-8542](https://issues.apache.org/jira/browse/CB-8542).

@shazron thoughts?

shazron:
Looks good

PR #83, opened 2015-04-06
solves CB-8804 https://github.com/apache/cordova-plugin-camera#android-quirks

thehuijb:
added action checkForSavedResult to be called from javascript in ondeviceready.

this pull request is an attempt to fix the following quirk:

Android Quirks

Android uses intents to launch the camera activity on the device to capture images, and on phones with low memory, the Cordova activity may be killed. In this scenario, the image may not appear when the Cordova activity is restored.


thehuijb:
Anyone had a chance to look at this yet?
I would appreciate any comments that help improve this PR

agrieve:
just went to have a look but the diff has changes that you didn't make in it. Can you try rebasing?

thehuijb:
alright, rebased to master, should be only my changes now.
sorry about that, it still git's me sometimes

src/android/CameraLauncher.java
```diff
@@ -163,7 +167,15 @@ else if ((srcType == PHOTOLIBRARY) || (srcType == SAVEDPHOTOALBUM)) {
             PluginResult r = new PluginResult(PluginResult.Status.NO_RESULT);
             r.setKeepCallback(true);
             callbackContext.sendPluginResult(r);
-            
+
+            return true;
+        } else if (action.equals("checkForSavedResult")) {
+            this.imageUri = Uri.fromFile(createCaptureFile(JPEG));
```
agrieve: should you do this only when savedRequestCode > 0?

src/android/CameraLauncher.java
```diff
@@ -163,7 +167,15 @@ else if ((srcType == PHOTOLIBRARY) || (srcType == SAVEDPHOTOALBUM)) {
...
+            if (savedRequestCode > 0) {
+                onActivityResult(savedRequestCode, savedResultCode, savedIntent);
+            } else {
+                callbackContext.success("no-result");
```
agrieve: Would be much nicer if returned result was false when coerced to a boolean (so that they can do if (result)). So maybe return empty string or null.

src/android/CameraLauncher.java
```diff
@@ -632,73 +645,79 @@ else if (destType == FILE_URI || destType == NATIVE_URI) {
...
     public void onActivityResult(int requestCode, int resultCode, Intent intent) {
 
-        // Get src and dest types from request code for a Camera Activity
-        int srcType = (requestCode / 16) - 1;
-        int destType = (requestCode % 16) - 1;
-
-        // If Camera Crop
-        if (requestCode >= CROP_CAMERA) {
-            if (resultCode == Activity.RESULT_OK) {
+        if (this.imageUri == null && this.getPictureFromGallery == false) {
```
agrieve: Would it be more appropriate to check if `callbackContext == null`? An app can be evicted even when getting a picture from the gallery.

www/Camera.js
```diff
@@ -72,4 +72,8 @@ cameraExport.cleanup = function(successCallback, errorCallback) {
     exec(successCallback, errorCallback, "Camera", "cleanup", []);
 };
 
+cameraExport.checkForSavedResult = function(successCallback, errorCallback) {
```
agrieve: Please update the plugin README.md with docs on this new API

src/android/CameraLauncher.java
thehuijb: right , should also do this when savedResultCode > 0

thehuijb:
Implemented changes based on @agrieve's input, could you have another look?

thehuijb:
@agrieve could you please have another look?

thehuijb:
forget about it

PR #85, opened 2015-04-15
Fix Android images opening in document viewer

dortzur:
When allowEdit is false, android opens the document viewer instead of photo gallery.
 The change uses ACTION_PICK in both cases, fixing the issue.


vladimir-kotikov:
Shouldn't we use `ACTION_GET_CONTENT` in both cases instead of `ACTION_PICK`?

@dortzur, could you please rebase you changes before merging.
Also before we can proceed with merging thi, could you please file it as described in [CONTRIBUTING.md](https://github.com/apache/cordova-plugin-camera/blob/master/CONTRIBUTING.md)

dortzur:
Nah, I'll probably just wait it out until I can move to React Native.
Thanks anyway.

PR #88, opened 2015-04-18
CB-8404 Rotate camera feed with device orientation

muratsu:
Fixed 2 bugs in windows where we have trouble rotating the image on preview and saving

### DIFF
--- a/src/windows/CameraProxy.js
+++ b/src/windows/CameraProxy.js
@@ -289,6 +289,17 @@ function takePictureFromCameraWP(successCallback, errorCallback, args) {
         captureSettings = null,
         CaptureNS = Windows.Media.Capture;
 
+    function cameraPreviewOrientation(arg) {
+        // Rotate the cam since WP8.1 MediaCapture outputs video stream rotated 90° CCW
+        if (screen.msOrientation === "portrait-primary" || screen.msOrientation === "portrait-secondary") {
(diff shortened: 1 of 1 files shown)


src/windows/CameraProxy.js
```diff
@@ -310,29 +319,50 @@ function takePictureFromCameraWP(successCallback, errorCallback, args) {
 
         tempFolder.createFileAsync(fileName, generateUniqueCollisionOption)
             .then(function(tempCapturedFile) {
-                capturedFile = tempCapturedFile;
-                return capture.capturePhotoToStorageFileAsync(encodingProperties, capturedFile);
+                return new WinJS.Promise(function (complete) {
+                    var imgStream = new Windows.Storage.Streams.InMemoryRandomAccessStream();
```
nikhilkh: imgStream.close() needs to be invoked once you are done using it to release system resources. It needs to be part of the success and error case.

nikhilkh: Same needs to be done for `fileStream`.

nikhilkh: Some comments on why the rotation needs to be done will be good - since its not quite obvious.

nikhilkh:
It might be a good idea to file a JIRA to improve the code for resizeImage and not to use canvas for a resize operation.

muratsu:
Rebased and squashed - ready to merge.

PR #91, opened 2015-04-30
[Android] plugin now returns and saves cropped image when allowEdit = true and destination is File or Native

fu-raz:
- Cropping a photo now actually returns the cropped image when saved to file
- Renamed getPicutresPath to getPicturesPath
- Trimmed trailing spaces


vladimir-kotikov:
@fu-raz, is this still actual? It looks like similar change was made already as a part of https://github.com/apache/cordova-plugin-camera/pull/120

fu-raz:
I have no idea, it was when I posted at the time. But I guess if it was fixed in #120, there's no need for this anymore.

vladimir-kotikov:
Okay, double-checked, the issue is fixed in master. @fu-raz, please close this PR.

PR #93, opened 2015-05-07
[Android] fix: support lollipop and earlier android versions to return a file path from uri regardless of file type.

SecuriTeam:
I've only tested this on Android versions:
LG Optimus L3 II - 4.1.1 - API 16
Samsung Galaxy S3 - 4.2.2 - API 17
HTC Evo - 4.3 - API 18
Nexus 4 - 5.0.0 - API 21
on Genymotion emulator.
and Galaxy Note 3 - 5.0.0 - API 21 on my actual phone.

### DIFF
--- a/src/android/FileHelper.java
+++ b/src/android/FileHelper.java
@@ -25,6 +25,9 @@ import android.os.Build;
 import android.provider.DocumentsContract;
 import android.provider.MediaStore;
 import android.webkit.MimeTypeMap;
+import android.util.Log;
+import android.os.Environment;
+import android.content.ContentUris;
 
 import org.apache.cordova.CordovaInterface;
 import org.apache.cordova.LOG;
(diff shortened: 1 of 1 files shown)


devpaul:
Hi, I was wondering if this was going to be reviewed soon. I'm receiving content uris for some media types on a Nexus 5 - 5.1 and I am hoping this could help resolve the issue.

infil00p:
Does this have a corresponding JIRA issue? Also, I'm going to talk about this on the dev list before pulling it in.

vladimir-kotikov:
@infil00p, is there any updates on this?

infil00p:
I've added a smaller fix in the Marshmallow camera to provide a real path, but it doesn't work with remote files.  We're not accepting this code in the repository without an ICLA or a CCLA, depending in how big SecuriTeam is.

PR #94, opened 2015-05-12
CB-8879 fix stripe issue with correct aspect ratio

muratsu:
This fixes the stripe issue we had on windows phone 8+. 
The fix was to set up resolutions with a common aspect ratio. Default aspect ratio is 1.78 (hd).

### DIFF
--- a/src/windows/CameraProxy.js
+++ b/src/windows/CameraProxy.js
@@ -57,6 +57,9 @@ module.exports = {
 var windowsVideoContainers = [".avi", ".flv", ".asx", ".asf", ".mov", ".mp4", ".mpg", ".rm", ".srt", ".swf", ".wmv", ".vob"];
 var windowsPhoneVideoContainers =  [".avi", ".3gp", ".3g2", ".wmv", ".3gp", ".3g2", ".mp4", ".m4v"];
 
+// Default aspect ratio 1.78 (16:9 hd video standard)
+var DEFAULT_ASPECT_RATIO = '1.8';
+
 // Resize method
 function resizeImage(successCallback, errorCallback, file, targetWidth, targetHeight, encodingType) {
     var tempPhotoFileName = "";
@@ -367,9 +370,8 @@ function takePictureFromCameraWP(successCallback, errorCallback, args) {
                 return;
             }
 
-            // Default aspect ratio 1.78 (16:9 hd video standard)
-            if (aspectRatios.indexOf("1.8") > -1) {
-                return setAspectRatio(capture, "1.8");
+            if (aspectRatios.indexOf(DEFAULT_ASPECT_RATIO) > -1) {
+                return setAspectRatio(capture, DEFAULT_ASPECT_RATIO);
             } else {
                 // Doesn't support 16:9 - pick next best
                 return setAspectRatio(capture, aspectRatios[0]);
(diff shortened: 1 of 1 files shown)


src/windows/CameraProxy.js
```diff
@@ -438,6 +457,61 @@ function takePictureFromCameraWP(successCallback, errorCallback, args) {
...
+        var aspectObj = allAspectRatios.reduce(function (map, item) {
+            if (!map[item]) {
+                map[item] = 0;
+            }
+            map[item]++;
+            return map;
+        }, {});
+
+        return Object.keys(aspectObj).filter(function (k) {
+            return aspectObj[k] === 3;
```
vladimir-kotikov: You are restricting aspect ratios to those which only supported by __all__ capture modes (picture, video and videoPreview). Not sure if it is by intent.

Probably it makes sense to respect __current__ capture mode and return aspect ratios for this capture mode. Potentially there could be a situation when device's camera support a different aspect ratios for video and image capture.

For example video is captured only in 640*480, which is equal to 1.3 aspect ratio, and for image there could be a number of other aspect ratios). This logic will return only '1.3' or even an empty set.

src/windows/CameraProxy.js
```diff
@@ -438,6 +457,61 @@ function takePictureFromCameraWP(successCallback, errorCallback, args) {
             });
     };
 
+    var getAspectRatios = function (capture) {
+        var photoAspectRatios = capture.videoDeviceController.getAvailableMediaStreamProperties(Windows.Media.Capture.MediaStreamType.photo).map(function (element) {
+            return (element.width / element.height).toFixed(1);
+        }).filter(function (element, index, array) { if (index === array.indexOf(element)) return 1; return 0; });
```
vladimir-kotikov: nit: it should be enough simply return result of comparison from filter function, like `return index === array.indexOf(element)`

src/windows/CameraProxy.js
```diff
@@ -438,6 +457,61 @@ function takePictureFromCameraWP(successCallback, errorCallback, args) {
...
+
+        return Object.keys(aspectObj).filter(function (k) {
+            return aspectObj[k] === 3;
+        });
+    };
+
+    var setAspectRatio = function (capture, aspect) {
+        // Max photo resolution with desired aspect ratio
+        var photoResolution = capture.videoDeviceController.getAvailableMediaStreamProperties(Windows.Media.Capture.MediaStreamType.photo).filter(function (elem) {
+            return ((elem.width / elem.height).toFixed(1) === aspect) ? 1 : 0;
```
vladimir-kotikov: nit: `return (elem.width / elem.height).toFixed(1) === aspect)` is sufficient for filtering, ternary operator is excess.

src/windows/CameraProxy.js
```diff
@@ -337,28 +336,48 @@ function takePictureFromCameraWP(successCallback, errorCallback, args) {
...
+
+            // Couldn't find a good ratio
+            if (aspectRatios.length === 0) {
+                destroyCameraPreview();
+                errorCallback('There\'s not a good aspect ratio available');
+                return;
+            }
+
+            // Default aspect ratio 1.78 (16:9 hd video standard)
+            if (aspectRatios.indexOf("1.8") > -1) {
```
vladimir-kotikov: IMO it will be better to use constant here instead: `DEFAULT_ASPECT_RATIO = 1.8`

robpaveza: jshint rule violation: should be single quotes.

muratsu: i need to add jshint later to this repo

src/windows/CameraProxy.js
muratsu: this is by design. all of the capture modes should be set to the same ratio otherwise there are artifacts on the captured image. this is a platform limitation.

PR #97, opened 2015-05-21
CB-8804 implement check for saved result

thehuijb:
Android uses intents to launch the camera activity on the device to capture images, and on phones with low memory, the Cordova activity may be killed.  
In this scenario, the image may not appear when the Cordova activity is restored.

Once our MainActivity restarts you should call the checkForSavedResult method in onDeviceReady to see if there is a result present from a previous call to getPicture.

### DIFF
--- a/README.md
+++ b/README.md
@@ -450,15 +450,18 @@ after calling `camera.getPicture`. Applies only when the value of
 ## navigator.camera.checkForSavedResult
 
 call this method in onDeviceReady to see if there is a result present from a previous call to getPicture.
-this in case 
+add the same options you used when calling getPicture, or leave it out to pass default options
 
-    navigator.camera.checkForSavedResult( successCallback, errorCallback );
+    navigator.camera.checkForSavedResult( successCallback, errorCallback, options );
 
 ### Description
 
 Tries to get the intermediate image file that is kept in temporary storage after calling `camera.getPicture`.
 Applies when the operating system decides to kill our MainActivity while getting an Image. 
-Once our MainActivity restarts you should call this method to see if a result is available
+Once our MainActivity restarts you should call this method to see if a result is available.
+
+If you want to use the same options you used when calling getPicture you have to make sure you store the options first,
+this way you can use your saved options to call checkForSavedResult, so the picture get's handled the same as it would have when Android hadn't killed our MainActivity
 
 ### Supported Platforms
 
@@ -466,7 +469,7 @@ Once our MainActivity restarts you should call this method to see if a result is
 
 ### Example
 
-    navigator.camera.checkForSavedResult(onSuccess, onFail);
+    navigator.camera.checkForSavedResult(onSuccess, onFail, savedOptions);
 
     function onSuccess(result) {
         console.log("checkForSavedResult success.");
--- a/README.md
+++ b/README.md
@@ -145,6 +145,7 @@ scenario, the image may not appear when the cordova activity is restored.
 Android uses intents to launch the camera activity on the device to capture
 images, and on phones with low memory, the Cordova activity may be killed.  In this
 scenario, the image may not appear when the Cordova activity is restored.
+You may try to get the image using the checkForSavedResult method. 
 
 #### Browser Quirks
 
@@ -440,6 +441,36 @@ after calling `camera.getPicture`. Applies only when the value of
 
     function onSuccess() {
         console.log("Camera cleanup success.")
+    }
+
+    function onFail(message) {
+        alert('Failed because: ' + message);
+    }
+
+## navigator.camera.checkForSavedResult
+
+call this method in onDeviceReady to see if there is a result present from a previous call to getPicture.
+this in case 
+
+    navigator.camera.checkForSavedResult( successCallback, errorCallback );
+
+### Description
+
+Tries to get the intermediate image file that is kept in temporary storage after calling `camera.getPicture`.
+Applies when the operating system decides to kill our MainActivity while getting an Image. 
+Once our MainActivity restarts you should call this method to see if a result is available
+
+### Supported Platforms
+
+- Android
+
+### Example
+
+    navigator.camera.checkForSavedResult(onSuccess, onFail);
+
+    function onSuccess(result) {
+        console.log("checkForSavedResult success.");
+        doSomethingWith(result);
     }
 
     function onFail(message) {
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ scenario, the image may not appear when the cordova activity is restored.
 Android uses intents to launch the camera activity on the device to capture
 images, and on phones with low memory, the Cordova activity may be killed.  In this
 scenario, the image may not appear when the Cordova activity is restored.
-You may try to get the image using the checkForSavedResult method. 
+You should try to get the image using the checkForSavedResult method. 
 
 #### Browser Quirks
 
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -127,16 +127,9 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
             this.mediaType = PICTURE;
             this.mQuality = 80;
 
-            this.mQuality = args.getInt(0);
-            destType = args.getInt(1);
-            srcType = args.getInt(2);
-            this.targetWidth = args.getInt(3);
-            this.targetHeight = args.getInt(4);
-            this.encodingType = args.getInt(5);
-            this.mediaType = args.getInt(6);
-            this.allowEdit = args.getBoolean(7);
-            this.correctOrientation = args.getBoolean(8);
-            this.saveToPhotoAlbum = args.getBoolean(9);
+            destType = args.optInt(1, FILE_URI);
+            srcType = args.optInt(2, CAMERA);
+            processConfiguration(args);
 
             // If the user specifies a 0 or smaller width/height
             // make it -1 so later comparisons succeed
@@ -171,7 +164,11 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
         } else if (action.equals("checkForSavedResult")) {
             this.imageUri = Uri.fromFile(createCaptureFile(JPEG));
             if (savedRequestCode > 0 || savedResultCode > 0) {
+                processConfiguration(args);
                 onActivityResult(savedRequestCode, savedResultCode, savedIntent);
+                savedRequestCode = 0;
+                savedResultCode = 0;
+                savedIntent = null;
             } else {
                 callbackContext.success("");
             }
@@ -183,6 +180,20 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
     //--------------------------------------------------------------------------
     // LOCAL METHODS
     //--------------------------------------------------------------------------
+
+    private void processConfiguration(JSONArray args) throws JSONException {
+        if (args == null)
+            throw new JSONException("no configuration object passed");
+        this.mQuality = args.optInt(0, 80);
+        // args 1 and 2 (destType, srcType) are skipped because they have no field representation in this class
+        this.targetWidth = args.optInt(3);
+        this.targetHeight = args.optInt(4);
+        this.encodingType = args.optInt(5, JPEG);
+        this.mediaType = args.optInt(6, PICTURE);
+        this.allowEdit = args.optBoolean(7);
+        this.correctOrientation = args.optBoolean(8);
+        this.saveToPhotoAlbum = args.optBoolean(9);
+    }
 
     private String getTempDirectoryPath() {
         File cache = null;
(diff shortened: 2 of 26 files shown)


stevengill:
@thehuijb hey. I will need to review this and see if it has any impact for other platforms before I can merge it in.

nikhilkh:
I think we should wait for @purplecabbage's design to improve the resume behavior for apps. This API looks very leaky and specific to Android - there is a similar scenario in Windows and it would be nice to tackle it fully.

stevengill:
Sounds good. Thanks for the input @nikhilkh

thehuijb:
@stevengill, I would appreciate the review.
@nikhilkh, this API is indeed specific to Android but I don't know what you mean by "it looks very leaky", doesn't feel leaky to me.

I think it would give the developers a means to handle this particular situation without to much trouble.
And although @purplecabbage is most likely going to produce a wonderful solution across the board (no pressure ;) it isn't quite there yet.

dpolivy:
I am looking at porting this into my fork of the camera plugin, and have a few suggestions/improvements to add (some I've addressed, some not yet). I'm not 100% done yet, but here's what I've seen off the bat:

* After retrieving the saved result, you should zero/null out the saved* instance variables.
* Creating this.imageUri should only happen inside the `if (savedRequestCode > 0...` block.
* The `onActivityResult` code can be simplified to just `return` after saving the values, to avoid the extra `else` block around the existing method.
* The `encodingType` is assumed to be JPEG when creating the temp file. This may not be the user's desired encoding. Is it possible to save some of the instance variables away in `onSaveInstanceState()`?
* Related to the prior comment, it might make sense for `checkForSavedResult` to take the same CameraOptions object that was passed to the original call, so it can properly restore state in the object.

I've updated my Nexus 5 to Android 5.1.1, and with the "always destroy activities" developer setting enabled, sometimes I'm finding that my main activity isn't being resumed at all -- after the camera disappears, I am left on the device home screen. Anyone else seen this?

thehuijb:
thanks @dpolivy, 

that's exactly the kind of review I was looking to get.

README.md
```diff
@@ -456,3 +457,36 @@ after calling `camera.getPicture`. Applies only when the value of
     function onFail(message) {
         alert('Failed because: ' + message);
     }
+
+## navigator.camera.checkForSavedResult
+
+call this method in onDeviceReady to see if there is a result present from a previous call to getPicture.
```
dpolivy: Capitalize first word

README.md
```diff
@@ -456,3 +457,36 @@ after calling `camera.getPicture`. Applies only when the value of
...
+add the same options you used when calling getPicture, or leave it out to pass default options
```
dpolivy: Capitalize first word. Also, options are required, but you can pass an empty object to get the defaults.

README.md
```diff
@@ -456,3 +457,36 @@ after calling `camera.getPicture`. Applies only when the value of
...
+
+    navigator.camera.checkForSavedResult( successCallback, errorCallback, options );
```
dpolivy: Minor: the `getPicture` examples in the README use `onSuccess` and `onFail` as names for the parameters. And no spacing after parens.

README.md
```diff
@@ -456,3 +457,36 @@ after calling `camera.getPicture`. Applies only when the value of
...
+
+### Description
+
+Tries to get the intermediate image file that is kept in temporary storage after calling `camera.getPicture`.
+Applies when the operating system decides to kill our MainActivity while getting an Image. 
+Once our MainActivity restarts you should call this method to see if a result is available.
+
+If you want to use the same options you used when calling getPicture you have to make sure you store the options first,
+this way you can use your saved options to call checkForSavedResult, so the picture get's handled the same as it would have when Android hadn't killed our MainActivity
```
dpolivy: No apostrophe in "gets"

src/android/CameraLauncher.java
```diff
@@ -124,16 +127,9 @@ public boolean execute(String action, JSONArray args, CallbackContext callbackCo
             this.mediaType = PICTURE;
             this.mQuality = 80;
```
dpolivy: I would get rid of all of these defaults here. I like your use of `optInt`, and passing in defaults; would suggest just eliminating the forced defaults above and go with the single line initialization in the method.

FWIW, here's how I ported this to my local copy:

        if (action.equals("takePicture")) {
            int srcType = args.optInt(2, CAMERA);
            int destType = args.optInt(1, FILE_URI);

            // Process the rest of the configuration arguments
            processConfiguration(args);
            ...

src/android/CameraLauncher.java
```diff
@@ -163,7 +159,19 @@ else if ((srcType == PHOTOLIBRARY) || (srcType == SAVEDPHOTOALBUM)) {
...
+        } else if (action.equals("checkForSavedResult")) {
+            if (savedRequestCode > 0 || savedResultCode > 0) {
+                processConfiguration(args);
+                imageUri = Uri.fromFile(createCaptureFile(encodingType));
+                onActivityResult(savedRequestCode, savedResultCode, savedIntent);
+                savedRequestCode = 0;
+                savedResultCode = 0;
+                savedIntent = null;
+            } else {
+                callbackContext.success("");
```
dpolivy: I'm not sure it's expected to get an empty string returned on success -- I believe the code will always return the requested result (URI to file, or base64 data). Better to just return an error here?

                callbackContext.error("No saved results available.");

src/android/CameraLauncher.java
```diff
@@ -173,6 +181,20 @@ else if ((srcType == PHOTOLIBRARY) || (srcType == SAVEDPHOTOALBUM)) {
...
+        this.mQuality = args.optInt(0, 80);
+        // args 1 and 2 (destType, srcType) are skipped because they have no field representation in this class
+        this.targetWidth = args.optInt(3);
+        this.targetHeight = args.optInt(4);
+        this.encodingType = args.optInt(5, JPEG);
+        this.mediaType = args.optInt(6, PICTURE);
+        this.allowEdit = args.optBoolean(7);
+        this.correctOrientation = args.optBoolean(8);
+        this.saveToPhotoAlbum = args.optBoolean(9);
+    }
```
dpolivy: Nice! FWIW, I added in defaults for all values. Here's my edit:

        this.mQuality = args.optInt(0, 80);
        // args 1 and 2 (destType, srcType) are skipped because they have no field representation in this class
        this.targetWidth = args.optInt(3, 0);
        this.targetHeight = args.optInt(4, 0);
        this.encodingType = args.optInt(5, JPEG);
        this.mediaType = args.optInt(6, PICTURE);
        this.allowEdit = args.optBoolean(7, false);
        this.correctOrientation = args.optBoolean(8, false);
        this.saveToPhotoAlbum = args.optBoolean(9, false);

www/Camera.js
```diff
@@ -72,4 +72,28 @@ cameraExport.cleanup = function(successCallback, errorCallback) {
...
+    var encodingType = getValue(options.encodingType, Camera.EncodingType.JPEG);
+    var mediaType = getValue(options.mediaType, Camera.MediaType.PICTURE);
+    var allowEdit = !!options.allowEdit;
+    var correctOrientation = !!options.correctOrientation;
+    var saveToPhotoAlbum = !!options.saveToPhotoAlbum;
+    var popoverOptions = getValue(options.popoverOptions, null);
+    var cameraDirection = getValue(options.cameraDirection, Camera.Direction.BACK);
+
+    var args = [quality, destinationType, sourceType, targetWidth, targetHeight, encodingType,
+                mediaType, allowEdit, correctOrientation, saveToPhotoAlbum, popoverOptions, cameraDirection];
```
dpolivy: It would be much better to refactor this as it's shared code with the getPicture API. Here's a refactor:

````
// Put this before the getPicture method definition

var prepareArgs = function(options) {
    options = options || {};
    var getValue = argscheck.getValue;

    var quality = getValue(options.quality, 50);
    var destinationType = getValue(options.destinationType, Camera.DestinationType.FILE_URI);
    var sourceType = getValue(options.sourceType, Camera.PictureSourceType.CAMERA);
    var targetWidth = getValue(options.targetWidth, -1);
    var targetHeight = getValue(options.targetHeight, -1);
    var encodingType = getValue(options.encodingType, Camera.EncodingType.JPEG);
    var mediaType = getValue(options.mediaType, Camera.MediaType.PICTURE);
    var allowEdit = !!options.allowEdit;
    var correctOrientation = !!options.correctOrientation;
    var saveToPhotoAlbum = !!options.saveToPhotoAlbum;
    var popoverOptions = getValue(options.popoverOptions, null);
    var cameraDirection = getValue(options.cameraDirection, Camera.Direction.BACK);

    return [quality, destinationType, sourceType, targetWidth, targetHeight, encodingType,
                mediaType, allowEdit, correctOrientation, saveToPhotoAlbum, popoverOptions,
                cameraDirection];
};

...

cameraExport.getPicture = function(successCallback, errorCallback, options) {
    argscheck.checkArgs('fFO', 'Camera.getPicture', arguments);
    var args = prepareArgs(options);

    exec(successCallback, errorCallback, "Camera", "takePicture", args);
    // XXX: commented out
    //return new CameraPopoverHandle();
};

...

cameraExport.checkForSavedResult = function(successCallback, errorCallback, options) {
    argscheck.checkArgs('fFO', 'Camera.checkForSavedResult', arguments);
    var args = prepareArgs(options);

    exec(successCallback, errorCallback, "Camera", "checkForSavedResult", args);
};
````

dpolivy:
@thehuijb Thanks for quickly reworking based on my feedback! I "ported" these into my fork and it seems to be working fine for me at the moment. I did add a bunch of additional comments directly on the diffs (sorry for the spam).

gregavola:
Hey @dpolivy - I'm testing this out on my S4 device and what I'm doing is:

- Enabled Do Not Keep Activities
- Then taking a photo and clicking Save
- Then when it brings it back to my App (which has been closed)
- After splash screen, it brings it do a blank White Screen, and deviceReady doesn't fire.

Is this the correct way to test?

dpolivy:
@gregavola That's roughly the process I'm using. What version of cordova-android are you using? It sounds like your app isn't loading -- can you confirm the HTML and JS are loading successfully? Anything useful from `adb logcat`?

gregavola:
The HTML/JS are working fine, but device ready never fires when the app re-loaded after the picture taken. Nothing really in logcat, it just stops when it reloads. The JS/HTML are there, just no ```deviceReady``` event. 

Here is the ```logcat```

```06-21 09:56:04.566: W/ApplicationPackageManager(25832): getCSCPackageItemText()
06-21 09:56:04.576: E/XXXX(25832): Intent Type null
06-21 09:56:04.576: E/XXXX(25832): Intent Action android.intent.action.MAIN
06-21 09:56:04.576: D/CordovaActivity(25832): Setting integer properties in CordovaActivity will be deprecated in 3.0 on July 2013, please use config.xml
06-21 09:56:04.576: D/CordovaActivity(25832): Setting boolean properties in CordovaActivity will be deprecated in 3.0 on July 2013, please use config.xml
06-21 09:56:04.586: I/CordovaLog(25832): Changing log level to ERROR(6)
06-21 09:56:04.596: D/CordovaWebView(25832): CordovaWebView is running on device made by: samsung
06-21 09:56:04.626: D/CordovaActivity(25832): Request code = 34
06-21 09:56:04.656: D/JsMessageQueue(25832): Dropping Native->JS message due to disabled bridge
06-21 09:56:04.947: D/JsMessageQueue(25832): Set native->JS mode to OnlineEventsBridgeMode
06-21 09:56:06.668: E/ViewRootImpl(25832): sendUserActionEvent() mView == null
06-21 09:56:07.049: I/chromium(25832): [INFO:async_pixel_transfer_manager_android.cc(56)] Async pixel transfers not supported
06-21 09:56:07.119: I/chromium(25832): [INFO:async_pixel_transfer_manager_android.cc(56)] Async pixel transfers not supported```

When I open up Chrome to inspect the element, there are no errors, and it just sits there. Any ideas?

Greg

thehuijb:
@gregavola, is there any way you could upgrade to the latest cordova or start using at least cordova-android 4.0.0.
there is a bug in lower versions that could be the cause of this.
so if possible please upgrade and try again.

dpolivy:
@gregavola It sounds like you have a broader Cordova issue. Perhaps [CB-7197](https://issues.apache.org/jira/browse/CB-7197)?

omefire:
Is the 'deviceReady' event supposed to be fired on resume ?

omefire:
Should we be exposing the 'checkForResult' function ? I'm thinking we should always handle this situation (intent gets killed behind the scenes) without the end user having to call any function or do anything. 
I suspect that might mean handling 'pause' and 'resume' events from within the plugin code itself. It might lead to a better abstraction.

thehuijb:
## How a cordova camera application works on android

This is basically  how a cordova application works:
image00
- User presses App-icon
- Applications start
- Mainactivity start
- Java side of plugins included
- WebView initialized
- Html loads
- javascript side of plugins loaded
- onDeviceReady fired

Now we have our application up and running.
It’s time to take a picture
image01

- We call getPicture
- The call goes to the CordovaBridge
- CordovaBridge uses the PluginManager to pass it to the camera Plugin
- Camera plugin launches a camera Activity

Now that a camera Activity is running, our MainActivity is pushed to the background, onPause is fired.

When the Camera Activity finished the MainActivity is called again via onActivityResult
image02

the result is dispatched to the javascript succesFunction where our application can then do with it as planned.
That’s what happens in a high memory availability scenario, a happy flow.

Now what happens when there isn’t enough memory left and the OS decides to kill our MainActivity?
image03

- CameraActivity is finished and sends its result back to the activity that launched it.
- Our MainActivity, is started.
- Java side of plugins included
- onActivityResult fires
- plugin fails to send result back to the javascript because it is not there anymore
- a completely new WebView is initialized
- Html loads
- javascript side of plugins loaded
- onDeviceReady fired

the precise order of event could differ on occasion but that doesn’t matter because the entire webview has to be rebuilt and with it all the javascript therefor the callback can never be completed in this situation.
You will always have to call a method in (or after) onDeviceReady to solve this problem.

Serge

src/android/CameraLauncher.java
thehuijb: this might be a little to much, optInt already defaults to 0 and optBoolean to false.
no need to explicitly define these defaults.

src/android/CameraLauncher.java
thehuijb: one should only return an error when an error occurs.
@agrieve advised me to callback success with empty string so you can handle the result in onSuccess javascript like so:
`if (result) {
  //do something
}`

adamduren:
+1 on this bugfix. I see this happen a lot.

gregavola:
Has this been integrated in the official plugin?

adamduren:
Since the pull request is still open I don't think so. I tried checking the release notes but they haven't been updated since 2013. :frowning:

stevengill:
Release notes are backwards. Oldest at top and newest at bottom. I know, it is weird.

adamduren:
ah, my bad. Thanks for pointing that out!

thehuijb:
@adamduren thanks for the +1

thehuijb:
@stevengill any chance you could review this?

sebastianor:
So, there is any way to resolve this problem ?

thehuijb:
@sebastianor, if you're asking if this pull request solves the issue, yes it does.
just call the checkForSavedResult method in onDeviceReady to see if there is a result present from a previous call to getPicture.

Bnaya:
@thehuijb this issue was solved in the platform level in
https://github.com/apache/cordova-android/commit/f5271431fbcda70188f485d9e91541bb596551fe
+
There's a plugin change developed to use the it

Its shame that they didn't took your PR sooner but now its redundant

thehuijb:
@Bnaya, thanks for the heads up. Don't know how I missed that, but great that this issue is finally tackled

Bnaya:
You can find more info here:
https://issues.apache.org/jira/browse/CB-8917

gregavola:
Does anyone have docs on how this works?

Greg

cenobyte321:
I'm using the 1.2.0 version of this plugin with cordova 5.4.1 and it crashes (doesn't reboot) the application if the memory is low and the main Cordova activity is not found. Applying the code from this pull request actually reboots the application. I don't fully understand how the fix @Bnaya posted is supposed to work but it's not behaving as expected... or am I missing something here?

Bnaya:
The fix in the cordova platform level is not released yet, unless you want to install the platform from GIT.
For the fixed plugin,
https://github.com/MSOpenTech/cordova-plugin-camera/commits/CB-9189
This is gonna be PR to the camera plugin that works with the platform change.

Until we will have all this, you can use fixes like this PR is offering, in the plugin level.
I have my own fix to that issue in the plugin + more issues fixes that you can see here:
https://github.com/photomania/cordova-plugin-camera/commits/prevent-crash-on-external-image-selection

cenobyte321:
Got it, thanks @Bnaya, I'll look then into using this PR temporarily until the final fix is released. The only thing I'm trying to figure out is, how can I save the state of other input fields the user entered in a form that are on the same page as the "capture photo" input? What comes to my mind is storing the state in Localstorage when the user selects to capture the photo and repopulate them in the onSuccess callback of the checkForSavedResult function. I don't know if there is a less "hacky" way to do it.

Bnaya:
The localstorge sounds like a good solution to me.
You can save there even more from the app state and restore the app to be as before the restart

nikhilkh:
@riknoll  Can you please provide details of your solution for this. We have fixed this in cordova-android and cordova-plugin-camera. It will require cordova-android 5.1 (soon to be released) and latest cordova-plugin-camera to fix this. Also, @riknoll has written docs on how to use this in your app.

riknoll:
As @nikhilkh mentioned, the new fix requires both `cordova-plugin-camera@2.1.0` and `cordova-android@5.1.0`. Updating to both of those won't automatically fix your application, you'll need to make some changes to handle the Android lifecycle as well. I've included a sample app that uses the new API as part of the Android Lifecycle Guide in the docs. It's published [here](http://cordova.apache.org/docs/en/dev/guide/platforms/android/lifecycle.html), but you should really check out [this PR](https://github.com/apache/cordova-docs/pull/461) as there are some pretty significant changes and a much more detailed sample app that haven't made it to the website yet. That doc should be useful for people not using the camera plugin as well because most of it applies to all apps targeting the Android platform. Plugin authors looking to use the new API should read the updated Android plugin guide [here](http://cordova.apache.org/docs/en/dev/guide/platforms/android/plugin.html). Hope this helps!

Titoine:
android@5.0.1 ? Can't upgrade to that version..
capture

Bnaya:
@Titoine 5.1.0

Titoine:
Sorry, mistake in my comment. But as you see on the screen I tried with 5.1.0

Bnaya:
Ok looks like it wan't published to NPM yet, just to git
https://www.npmjs.com/package/cordova-android
https://github.com/apache/cordova-android/releases

you should install it from github, or wait for the npm  release which should come soon (?)
https://github.com/apache/cordova-android.git#5.1.0

stevengill:
Released now
http://cordova.apache.org/announcements/2016/01/24/cordova-android-5.1.0.html

Titoine:
I'm testing it with the developper option "Don't keep activities" and this isn't working everytime.
After a camera camera getPicture, I get the 'resume' event, and the next build, it isn't launched anymore..only the deviceready.

My existing app isn't really prepare for a solution like this and It take time to rebuild everything (webview etc.) so I'm searching an alternative. If anybody have any idea, I'll take it.

There is a solution for camera part with the CordovaCameraPreview plugin (https://github.com/mbppower/CordovaCameraPreview). But I didn't find a solution for image selection in the gallery and the foreground gallery doesn't seem to work anymore (It crash with "Don't keep activities").

riknoll:
@Titoine you may be seeing a known issue in the fix (see [CB-10498](https://issues.apache.org/jira/browse/CB-10498)). Basically, it is possible in some circumstances to "miss" the resume event if you don't add event handlers in time. Let me know if you keep seeing the issue (and comment on the linked JIRA).

adamduren:
@Titoine I experienced this as well. I thought it was a device specific bug but it was resolved by attaching the listeners as soon as possible.

riknoll:
@Titoine @adamduren I've got a PR open in cordova-android that should fix the `resume` event issues. It should let you subscribe to the `resume` event after `deviceready` fires with no issue. Check it out here: 

https://github.com/apache/cordova-android/pull/257

I would love some feedback if you have a chance to try it out. I hope to merge it in soon.

Titoine:
@riknoll 
can't test it right now. I'm trying to do a camera/gallery plugin without having to use native and avoid this problem.

nikhilkh:
@thehuijb Can you please close this PR? It's no relevant anymore.

PR #99, opened 2015-05-28
Try to use realpath filename instead of default modified.jpg

nantunes:


schreyers:
Hi Nantunes, 

Did you manage to use the realpath filename and if so which version of the plugin is it in is this in? There is a current bug which means images get uploaded as duplicates because of the file name still being modified.jpg for all the edited images.

https://stackoverflow.com/questions/30438483/cordova-image-upload-duplicate-issue#

I have recently re-added the plugin but I still get the issue that edited images are called modified.jpg.

Thanks,
Jon

nantunes:
Hi @schreyers 

I'm still using my forked version, as this PR hasn't been merged (or even reviewed).

If it is critical to your project, you probably better applying this fix yourself.

schreyers:
Thanks @nantunes 

I'll make your changes to the files and remove the source="npm" from my config file.

schreyers:
Hi @nantunes 

Quick question. Are you able to use your forked version when using build.phonegap.com? If not how are you using your changes?

Thanks in advance,

nantunes:
I'm using the lastest versions of cordova with ionic framework. I build locally. No phonegap-build, so I can't tell.

schreyers:
Sorry for the delayed reply. Have been trying to build locally but i get loads of errors. 

I really need this to be fixed in the npm version of the plugin.

src/android/CameraLauncher.java
```diff
@@ -498,15 +498,19 @@ private void refreshGallery(Uri contentUri)
...
         // Create an ExifHelper to save the exif data that is lost during compression
-        String modifiedPath = getTempDirectoryPath() + "/modified.jpg";
+        String modifiedPath = getTempDirectoryPath() + "/" + fileName;
 
         OutputStream os = new FileOutputStream(modifiedPath);
         bitmap.compress(Bitmap.CompressFormat.JPEG, this.mQuality, os);
         os.close();
 
-        // Some content: URIs do not map to file paths (e.g. picasa).
-        String realPath = FileHelper.getRealPath(uri, this.cordova);
```
vladimir-kotikov: These two lines above seems excess, no? You've got `realPath` already at the beginning of this method.

src/android/CameraLauncher.java
```diff
@@ -498,15 +498,19 @@ private void refreshGallery(Uri contentUri)
 
 
 private String ouputModifiedBitmap(Bitmap bitmap, Uri uri) throws IOException {
+        // Some content: URIs do not map to file paths (e.g. picasa).
+        String realPath = FileHelper.getRealPath(uri, this.cordova);
+
+        // Get filename from uri
+        String fileName = realPath != null ? realPath.substring(realPath.lastIndexOf('/') + 1) : "modified.jpg";
```
vladimir-kotikov: This doesn't respect `encodingType` parameter. In case if `encodingType == PNG` resultant filename still might have '.jpg' extension.

vladimir-kotikov:
@nantunes, thanks for contribution. Could you please rebase your changes.

Also please file an Apache CLA as described in [CONTRIBUTING.md](https://github.com/apache/cordova-plugin-camera/blob/master/CONTRIBUTING.md)

src/android/CameraLauncher.java
nantunes: These two lines were removed, not added...

src/android/CameraLauncher.java
nantunes: Just kept the same behaviour as before (see bellow old line 502). It would be nice to respect the encoding, but outside the scope of this patch.

nantunes:
Ok, will do it this night.

src/android/CameraLauncher.java
vladimir-kotikov: Sorry, my bad :)

nantunes:
Rebased and ICLA sent by email.

vladimir-kotikov:
Yep, see it now. Merging. Thanks for contribution, @nantunes!

FuzzyTree:
Has anyone tested this patch? Applying the changes to CameraLauncher.java results in failed `getPicture` calls. Unfortunately the only error is "Error retrieving image", so I'm not sure where it breaks.

I also tried building from the nightlies and get the same error. When I revert `ouputModifiedBitmap` in the nightlies the error goes away.

nantunes:
_Per se_ the patch only manipulates strings:
```java

String modifiedPath = getTempDirectoryPath() + "/" + fileName;
```

So I don't how could it cause that error. Maybe some other change in-between? Could you diff the two versions of CameraLauncher.java?

FuzzyTree:
Debugging shows that `FileHelper.getRealPath` returns an empty (non-null) string, which causes the function to return a directory name instead of a file name.

`/storage/emulated/0/Android/data/com.ionicframework.ws938141/cache: open failed: EISDIR (Is a directory)`

        // Some content: URIs do not map to file paths (e.g. picasa).
        // Get filename from uri

Adding a check for empty string fixed the issue for me

    String fileName = (realPath != null && !realPath.isEmpty()) ?

shraddha-khalekar:
that issue is still there in camera plugin

shraddha-khalekar:
can someone plz tell how to solve this issue for ionic 4

janpio:
Please open a new issue and fill the template instead of abusing an old and already merged PR.

PR #100, opened 2015-05-29
Update docs

muratsu:
Update docs to enhance reading experience


README.md
```diff
@@ -35,6 +35,16 @@ Although the object is attached to the global scoped `navigator`, it is not avai
 
     cordova plugin add cordova-plugin-camera
 
+## API
+- Camera
+    - navigator.camera.getPicture(success, fail, options)
```
nikhilkh: Is this like a TOC? Could this be pointing to anchor links below?

README.md
```diff
@@ -83,47 +93,68 @@ quality, even if a `quality` parameter is specified.  To avoid common
 memory problems, set `Camera.destinationType` to `FILE_URI` rather
 than `DATA_URL`.
 
-### Supported Platforms
+#### Supported Platforms
+
+![](doc/img/android-success.png) ![](doc/img/blackberry-success.png) ![](doc/img/browser-success.png) ![](doc/img/firefox-success.png) ![](doc/img/fireos-success.png) ![](doc/img/ios-success.png) ![](doc/img/windows-success.png) ![](doc/img/wp8-success.png) ![](doc/img/ubuntu-success.png)
```
nikhilkh: These seems hard to author - but easy to consume. For the authoring experience, I wonder if you have thought about some tooling experience.

README.md
muratsu: It can, but it won't work outside of GH. If that's not a problem I'm :+1:

nikhilkh: Eh...That's not ideal.

README.md
muratsu: I don't think we'll be changing these frequent enough to justify tooling.

nikhilkh: While this prettier, another caveat is that it not searchable any more. In any case, this is a good improvement we should commit this.

PR #104, opened 2015-07-02
CB-9193: Add 'showLibraryButton' to allow choosing source

dpolivy:
A new camera option, `showLibraryButton`, is added which allows the user
to easily switch from the camera to the photo library once they have
entered the plugin.

Currently, this is optimally supported on iOS (7+), with additional
implementations for Android and WP8 based on what was available for
those platforms. Please see the documentation updates (specifically, the 'Quirks' changes) for details on how this is implemented on each platform.

I realize that the behavior on Android and Windows Phone is not 100% consistent to iOS, but it is the best that can be done given the constraints of what's available on those platforms. If we were to rewrite this plugin to use completely custom camera UIs, we could do a whole lot better -- but that's a much bigger undertaking.

I also made some opportunistic changes while I was working on this, including cleaning up parameter initialization for Android, moving hardcoded strings to localizable ones, fixing some space/tab and indentation inconsistencies, and other code cleanup.

All modifications here do not impact the default functionality of someone who is currently using this plugin (though it would be great for additional test coverage on that). In order for this change to work, `showLibraryButton: true` must be passed in the options, and the source type must be `CAMERA`.

@infil00p @shazron @nikhilkh @stevengill Is it possible to get reviews for your respective platforms?

Fixes https://issues.apache.org/jira/browse/CB-9193

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,20 +71,20 @@ Documentation consists of template and API docs produced from the plugin JS code
 
 
 * [camera](#module_camera)
-  * [.getPicture(successCallback, errorCallback, options)](#module_camera.getPicture)
-  * [.cleanup()](#module_camera.cleanup)
-  * [.onError](#module_camera.onError) : <code>function</code>
-  * [.onSuccess](#module_camera.onSuccess) : <code>function</code>
-  * [.CameraOptions](#module_camera.CameraOptions) : <code>Object</code>
+    * [.getPicture(successCallback, errorCallback, options)](#module_camera.getPicture)
+    * [.cleanup()](#module_camera.cleanup)
+    * [.onError](#module_camera.onError) : <code>function</code>
+    * [.onSuccess](#module_camera.onSuccess) : <code>function</code>
+    * [.CameraOptions](#module_camera.CameraOptions) : <code>Object</code>
 
 
 * [Camera](#module_Camera)
-  * [.DestinationType](#module_Camera.DestinationType) : <code>enum</code>
-  * [.EncodingType](#module_Camera.EncodingType) : <code>enum</code>
-  * [.MediaType](#module_Camera.MediaType) : <code>enum</code>
-  * [.PictureSourceType](#module_Camera.PictureSourceType) : <code>enum</code>
-  * [.PopoverArrowDirection](#module_Camera.PopoverArrowDirection) : <code>enum</code>
-  * [.Direction](#module_Camera.Direction) : <code>enum</code>
+    * [.DestinationType](#module_Camera.DestinationType) : <code>enum</code>
+    * [.EncodingType](#module_Camera.EncodingType) : <code>enum</code>
+    * [.MediaType](#module_Camera.MediaType) : <code>enum</code>
+    * [.PictureSourceType](#module_Camera.PictureSourceType) : <code>enum</code>
+    * [.PopoverArrowDirection](#module_Camera.PopoverArrowDirection) : <code>enum</code>
+    * [.Direction](#module_Camera.Direction) : <code>enum</code>
 
 * [CameraPopoverHandle](#module_CameraPopoverHandle)
 * [CameraPopoverOptions](#module_CameraPopoverOptions)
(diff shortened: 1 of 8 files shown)


muratsu:
I would like to hear everyone's opinion on this before going with the code review.

Showing the photo gallery on bottom left seems like a custom view behavior for the stock Camera app. I've tried this with couple apps and none of them, except WhatsApp had similar behavior.

I think we should think about enabling custom camera UI scenarios instead of adding more platform quirks. But as noted it's a big undertaking.

Note: In WhatsApp nothing happens when you actually tap on the black thumbnail.

Slack


WhatsApp


Messages

dpolivy:
@muratsu Valid question, thanks for starting the discussion. The Android implementation is more like the Messages/Slack experience you highlight above. I think the "right" solution here really depends on the app and the desired user experience. For me, camera is the 90% usage scenario, but it's important to allow gallery upload, too. So the best UX is to launch right into the camera UI, with an option to fall out into the gallery. That creates an extra tap for only a smaller percentage of users, vs all users if we show a picker off the bat (e.g., like Android).

Supporting better custom camera UI longer term may be a better way to solve this, but there's no way to defer that to the app developer without requiring them to go build some native components.

To be honest, longer term I will likely fork this plugin, and replace the Android and Windows Phone implementations with ones using the low level camera APIs to better replicate the iOS approach, so I can build custom UIs on top of them that better match our usage scenarios. But, as a first (quicker) stab, I'm trying it this way, and wanted to give the opportunity to share this back as I can't imagine I'm the only one who would find this valuable.

I'm curious to hear other's thoughts on this. Adding a more consistent "picker" UI which can be implemented consistently on all platforms might be a viable alternative -- BUT it kills the use case I mentioned above (which was my motivation for doing it in the first place). Plus, that can sort of be done already today by having separate buttons in your HTML UI.

muratsu:
Ping

dpolivy:
@infil00p @shazron @nikhilkh @stevengill Any thoughts on this discussion or possibility of a CR?

jcesarmobile:
-1 to adding new quirks

infil00p:
-1 to this as well.  This wouldn't be consistent on Android, since we're using Intents to launch a camera, and it's possible some devices wouldn't have the library icon on their Camera activity.

dpolivy:
@jcesarmobile Is this a quirk just because it's only implemented on iOS, Android, and WP?

@infil00p Actually, it is consistent on Android, and has nothing to do with whether the device has a library icon on their Camera activity (I'm not sure that any do?). We're simply showing an Intent Chooser that lets the user choose which Activity they want to launch, and the list includes all of the camera and file activities found on their device. That is consistent across all Android devices. And it's using the exact same intents that would be used if you chose to show the camera or library. Give it a try and see the behavior for yourself.

I still think this feature is valuable and provides a better user experience in certain circumstances. We've had success using this in our app, and being a good citizen I've tried to make it available for others to take advantage of. There are some native limitations here, but if you truly wanted a consistent camera experience and UI across all platforms, you'd basically have to write to the raw camera APIs on each platform which is a significant departure from the current approach of this plugin.

purplecabbage:
-1
This should probably be 3 plugins
1. camera plugin
2. image picker plugin
3. image src chooser plugin

We can revisit this later, we are doing a review of many camera features currently ( plus overlap with media capture )

One other concern I have is the localization of text.  Where are the strings coming from currently?

infil00p:
I agree with @purplecabbage, and I'm still -1 on this.

dpolivy:
@purplecabbage On Android, the strings were previously [hardcoded](https://github.com/apache/cordova-plugin-camera/pull/104/files#diff-f2e83abdaf34884df7e74b6493cf2dd7L79) into the plugin. My commit actually [moved them out into plugin.xml](https://github.com/apache/cordova-plugin-camera/pull/104/files#diff-53f390d375398624afe1cfe1125f42bfR79) so they could actually be localized. The other strings (e.g. the intent names) are actually provided by the OS based on the available activities. For WP, it is all built in to the OS, and for iOS the one new string (Library) is added in a localizable manner consistent with other strings in plugins.

I don't necessarily disagree with your approach to splitting this into multiple plugins, however I really do believe there is a useful scenario where the camera UI launches first, but the user has a quick escape into the library. That wouldn't be supported by splitting into 3 plugins. Instead if you really want more of a consistent experience you just have to write to the native camera APIs directly on each platform and built a custom "Cordova" standard UI on top of them. Having a modal source chooser step just adds complexity to the user experience in most cases. Far better from a UX perspective to default to the 90% case (as determined by the developer for their app) but allow easy toggling between the various options.

In any case, it seems that there's no future hope for this PR in its current form. I still do advocate that the problem I'm trying to solve is a valuable and important addition to the Cordova plugins, but I'm certainly open to the form it takes as long as it meets the user experience bar we're trying to achieve.

PR #107, opened 2015-07-12
Update CameraLauncher.java

Jeff86:
Replace black background to white background for transparent pics from galery

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -55,6 +55,9 @@ import android.provider.MediaStore;
 import android.util.Base64;
 import android.util.Log;
 import android.content.pm.PackageManager;
+
+import android.graphics.Canvas;
+import android.graphics.Color;
 /**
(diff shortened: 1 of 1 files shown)


vladimir-kotikov:
@Jeff86, could you please explain the rationale of this change? For me it looks like unrelated to plugin functionality.

Jeff86:
@vladimir-kotikov thanks for your comment.

This plugin propose to take pictures from gallery, no-one wants an imported GIF with transparent background going black. We mainly use white as a standard replacement color for transparent layer and my update makes it fit the current standard.

The rational of this change is to fit current standards, make life easier for users and propose a tested and working way to handle a usual issue for developers.

vladimir-kotikov:
There are a number of issues with this:
* IMO the image editing is out of this plugin's scope
* This change solves your problem, however other devs might want to use images from gallery as-is.
* Image processing causes a major performance gap and leads to bad UX

Sorry, but I'm not sure we should merge this.

PR #110, opened 2015-07-27
CB-9151 Trigger captureAction only once

muratsu:
Prevent multiple captureAction calls by using `clicked` attribute


purplecabbage:
This seems to work, but perhaps removing the listener on the first call would be a better approach.
Also, I assume I can button mash the cancel button and get the same effect.

muratsu:
I think the logic has to stay in onClick event for easy cleanup purposes. I've made the event handlers seperate functions. It's much cleaner now.

PR #111, opened 2015-08-03
CB-9443 Pick correct maxResolution

muratsu:
According to MSDN docs we're currently not picking the correct maxResolution for images:
https://msdn.microsoft.com/en-us/library/windows/apps/windows.media.capture.cameracaptureuimaxphotoresolution.aspx

### DIFF
--- a/src/windows/CameraProxy.js
+++ b/src/windows/CameraProxy.js
@@ -696,21 +696,22 @@ function takePictureFromCameraWindows(successCallback, errorCallback, args) {
     // decide which max pixels should be supported by targetWidth or targetHeight.
     var maxRes = null;
     var UIMaxRes = WMCapture.CameraCaptureUIMaxPhotoResolution;
+    var totalPixels = targetWidth * targetHeight;
     switch (true) {
-        case (targetWidth >= 1280 || targetHeight >= 960) :
-            cameraCaptureUI.photoSettings.maxResolution = UIMaxRes.large3M;
+        case (totalPixels <= 320 * 240):
+            cameraCaptureUI.photoSettings.maxResolution = UIMaxRes.verySmallQvga;
             break;
-        case (targetWidth >= 1024 || targetHeight >= 768) :
-            maxRes = UIMaxRes.mediumXga;
-            break;
-        case    (targetWidth >= 800 || targetHeight >= 600) :
-            maxRes = UIMaxRes.mediumXga;
-            break;
-        case  (targetWidth >= 640 || targetHeight >= 480) :
+        case (totalPixels <= 640 * 480):
             maxRes = UIMaxRes.smallVga;
             break;
-        case    (targetWidth >= 320 || targetHeight >= 240) :
-            maxRes = UIMaxRes.verySmallQvga;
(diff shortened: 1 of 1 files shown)


src/windows/CameraProxy.js
```diff
@@ -696,21 +696,22 @@ function takePictureFromCameraWindows(successCallback, errorCallback, args) {
     // decide which max pixels should be supported by targetWidth or targetHeight.
     var maxRes = null;
     var UIMaxRes = WMCapture.CameraCaptureUIMaxPhotoResolution;
+    var totalPixels = targetWidth * targetHeight;
     switch (true) {
```
robpaveza: As long as you're refactoring the numbers, let's also turn this from `switch (true)` into `if`/`else`.

muratsu: didn't notice `switch (true)` was just correcting the case params but wow - haha. definetely turning this to a `if/else`

src/windows/CameraProxy.js
```diff
@@ -696,21 +696,22 @@ function takePictureFromCameraWindows(successCallback, errorCallback, args) {
...
+        case (totalPixels <= 3 * 1000 * 1000):
+            maxRes = UIMaxRes.large3M;
             break;
-        case    (targetWidth >= 320 || targetHeight >= 240) :
-            maxRes = UIMaxRes.verySmallQvga;
+        case (totalPixels <= 5 * 1000 * 1000):
+            maxRes = UIMaxRes.veryLarge5M;
             break;
         default :
             maxRes = UIMaxRes.highestAvailable;
```
robpaveza: Add `break;` after this line.

purplecabbage: switch(true) is faster, + Personally, I find it easier to read.

muratsu:
Sorry didn't see your comment earlier. 
In terms of perf, I think this the usage of `if/else` vs `switch` is not really important here. Since we're only executing it once, the perf difference is not going to be noticable either case.
Readability is a bit more subjective - I don't have a preference in this case. I tend to use switches only when I have fall-through cases. 
I wasn't familiar with `switch(true)` pattern. It seemed like an anti-pattern to me at first but looks like it is a common pattern in dynamic languages. I'll revert it back to switch.

rgecy:
How do I unsubscribe from this plugin list. I do not want to get emails on
every update.

src/windows/CameraProxy.js
robpaveza: switch(true) would only be faster in the event that your cases were constants and you were doing explicit equality checks (e.g., `case 10:`).  They're not, though, and because you're using greater-than comparisons, you lose the ability of the runtime compiler to optimize using jump-tables.  If we're really concerned about perf, I'd suggest adding profiling code to order the most-common to the least-common cases; but I think this code will run so infrequently, it's not a perf-critical path.

purplecabbage: Yes, the 'faster' argument is not particularly strong in this case
considering how often it will be called.



My team is hiring!
@purplecabbage
risingj.com

PR #112, opened 2015-08-04
CB-9446 Fix real path returning null pointer on some Samsung devices

charlesverge:
Fix for https://issues.apache.org/jira/browse/CB-9446

### DIFF
--- a/src/android/ExifHelper.java
+++ b/src/android/ExifHelper.java
@@ -53,7 +53,11 @@ public class ExifHelper {
      * @throws IOException
      */
     public void createInFile(String filePath) throws IOException {
-        this.inFile = new ExifInterface(filePath);
+        if (filePath != null) {
+            this.inFile = new ExifInterface(filePath);
+        } else {
+            throw new IOException("null pointer passed to createInFile");
+        }
     }
 
     /**
@@ -63,7 +67,11 @@ public class ExifHelper {
      * @throws IOException
      */
     public void createOutFile(String filePath) throws IOException {
-        this.outFile = new ExifInterface(filePath);
+        if (filePath != null) {
+            this.outFile = new ExifInterface(filePath);
+        } else {
+            throw new IOException("null pointer passed to createOutFile");
+        }
     }
 
     /**
--- a/src/android/FileHelper.java
+++ b/src/android/FileHelper.java
@@ -33,6 +33,7 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Locale;
+import android.util.Log;
 
 public class FileHelper {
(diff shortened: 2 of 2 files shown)


devteamshippeo:
I have the same issue on Samsung J1 this morning
Thanks for this fix !

riknoll:
Is this fixing the same issues as #93?

nikopolv:
Still crashing on Sony Aqua M2

charlesverge:
@nikopolv This fix usually covered pre Kit Kat devices, does #93 execute properly with your Sony Aqua M2 ? Also, can you indicate the os version you have on the device ?

charlesverge:
@riknoll The code in #93 is missing a null pointer check
But the differences for the code which retrieves the url is
#112 
+                Cursor cursor = cordova.getActivity().managedQuery(Uri.parse(uriString), proj, null, null, null);
+                int column_index = cursor.getColumnIndexOrThrow(_DATA);
+                cursor.moveToFirst();
+                realPath = cursor.getString(column_index);

#93
+			cursor = context.getContentResolver().query(uri, projection, selection, selectionArgs,
+					null);
+			if (cursor != null && cursor.moveToFirst()) {
+				final int column_index = cursor.getColumnIndexOrThrow(column);
+				return cursor.getString(column_index);
+			}

I don't have time scheduled atm to do device testing

nikopolv:
@charlesverge The Sony Aqua M2 had 4.4 kitkat running in it.

charlesverge:
@nikopolv Does the Sony Aqua M2 with 4.4 kitkat work with the branch issue #93 is based off of ?

nikmartin:
On a Samsung S5 with Android 5.0.1 this patch works fine

infil00p:
@charlesverge, do you have an ICLA signed for this? Usually when people write entire new methods, I need an ICLA before I can accept public contributions like this.  

You can find the info here: http://www.apache.org/licenses/#clas

BTW, I like this better than #93.

charlesverge:
@infil00p The method comes from a previous version of the plugin which was working for some devices.
https://github.com/apache/cordova-plugin-camera/blob/0d63ac27ca6911d34a3eef76e393d64e898d8b47/src/android/FileHelper.java

infil00p:
@charlesverge Yeah, and it was slated to be "deprecated" for some reason, which is why I forgot about it.  I added it back in since we can't use the ResourceAPI in the same way.

infil00p:
OK, do we still need this, or does the N fix also fix this? More testing is needed.

infil00p:
This probably got clobbered when I added a FileProvider to end all this getRealPath nonsense that was breaking everything.  I'm going to close this for now.

PR #113, opened 2015-08-04
Optimize for clarity

purplecabbage:
changed 'case' to else-if
added WinJS to JSHint global ignores


vladimir-kotikov:
@purplecabbage, is this still valid?

purplecabbage:
Yeah. Sorry, I thought I merged it.

PR #115, opened 2015-08-18
CB-9490 Fixed cleanup function

cojomojo:
Modfied cleanup function to deal with camera applications that save photos that should have been discarded when the retake option was chosen

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -422,7 +422,7 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
             }
 
             this.processPicture(bitmap, this.encodingType);
-            checkForDuplicateImage(DATA_URL);
+            cleanupTempImages(DATA_URL);
         }
 
         // If sending filename back
@@ -485,8 +485,8 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
             throw new IllegalStateException();
         }
 
-        this.cleanup(FILE_URI, this.imageUri, uri, bitmap);
         bitmap = null;
+        this.cleanup(FILE_URI, uri, bitmap);
     }
 
 private String getPicutresPath()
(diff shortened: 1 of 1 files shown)


infil00p:
You removed part of the cropping functionality, we can't accept this pull request as it is.

cojomojo:
That was not suppose to be removed, must have done that accidentally. Ill fix it, and update the PR in a few minutes.

cojomojo:
Fixed the PR to only affect the cleanup function. Please re review.

cojomojo:
Is this PR still being considered?

nikhilkh:
@sgrebnov to see if you can help merge this. Do we have the correct test coverage for this?

sgrebnov:
@cojomojo the fix makes perfect sense, but I see that the code you use is very similar to what we have in `checkForDuplicateImage` and looks like that function is not required anymore (as I think you cover that case) so I propose to rename `checkForDuplicateImage` to `removeTemporaryImages` (or `cleanupTempImages`) and move your logic there.

src/android/CameraLauncher.java
```diff
@@ -983,12 +983,39 @@ private Cursor queryImgDB(Uri contentStore) {
...
+
+            if (imageType == FILE_URI) {
+                id--;
+                for (int i = diff; i > 0; i--) {
+                    Uri uri = Uri.parse(contentStore + "/" + id);
+                    this.cordova.getActivity().getContentResolver().delete(uri, null, null);
+                    id--;
+                }
+            } else {
+                for (int i = diff; i > 0; i--) {
```
sgrebnov: looks like this `for` loop is the same as above; what if we keep only one loop, for example

```
if (imageType == FILE_URI) {
    id--;
}

for (int i = diff; i > 0; i--) {
    Uri uri = Uri.parse(contentStore + "/" + id);
    this.cordova.getActivity().getContentResolver().delete(uri, null, null);
    id--;
}
```

cojomojo:
@sgrebnov Ok, I finally got around to doing what you suggested. Please re-review.

src/android/CameraLauncher.java
```diff
@@ -241,6 +241,7 @@ public void takePicture(int returnType, int encodingType) {
      * @return a File object pointing to the temporary picture
      */
     private File createCaptureFile(int encodingType) {
+        File photo = null;
```
sgrebnov: Looks like this variable is not used

src/android/CameraLauncher.java
```diff
@@ -386,12 +387,10 @@ private void processResultFromCamera(int destType, Intent intent) throws IOExcep
             FileHelper.stripFileProtocol(this.croppedUri.toString()) :
             FileHelper.stripFileProtocol(this.imageUri.toString());
 
-        if (this.encodingType == JPEG) {
-            try {
-                //We don't support PNG, so let's not pretend we do
-                exif.createInFile(sourcePath);
-                exif.readExifData();
-                rotate = exif.getOrientation();
+            //We don't support PNG, so let's not pretend we do
```
sgrebnov: I don't think this change should be in scope of this PR (I know there is a person working on .PNG support so this will just bring merge conflict)

sgrebnov:
Reviewed: :+1: thank you @cojomojo!
@vladimir-kotikov, I know that you are working on some other improvements so please merge this along with your work!

cojomojo:
@sgrebnov Something got messed up when I merged in the latest before resubmitting. I doubt this is even valid Java right now, I'll fix it later today and then we should be good to go.

infil00p:
Can this wait until after my Cordova-Android 5.0.x pull request is merged in? checkForDuplicateImage requires a permission on Android 6.0+ and I want to make sure that we don't accidentally add something without a permissions check.

cojomojo:
@infil00p Good idea, let me know when it's merged

cojomojo:
@sgrebnov @vladimir-kotikov OK, I fixed the issue caused by a bad merge. This PR is ready once we see how @infil00p 's PR for Cordova-Android 5.0.x  and the permissions issue he mentioned affects things.

infil00p:
@cojomojo Won't this still have the problem you found when you filed this? https://issues.apache.org/jira/browse/CB-9506

cojomojo:
@infil00p Yes, it would still have the problem. I submitted this fix still because it is trivial (this PR won't change how the plugin is used at all, and only changes 2 lines, outside of the function itself), while the fix for 9506 will be non-trivial.  Therefore, I think that unless 9506 is going to be fixed soon we should still patch 9490 with this.

cojomojo:
@infil00p Should we close this PR? After thinking about this for quite some time, I think fixing https://issues.apache.org/jira/browse/CB-9506 would be favorable over this and would by consequence make this PR irrelevant anyways.

PR #117, opened 2015-08-26
feat(ios): Manual dismissal method (CB-9980)

sarahgoldman:
I have added a new function called `dismiss` that allows one to dismiss any camera popover views and view controllers on iOS. I encountered a need for this function when implementing a timeout based logout on my app. If the user timed out while sitting there with the camera open, I had no way to close the camera view controller and it would stay open on top of the login page. When they finally took a photo it would exit the camera and they would be on the login page. Now I can call `navigator.camera.dismiss()` before I go to the login view.

`dismiss` does not take any callback handlers, instead it triggers the error callback set with `getPicture` and sends a message stating the camera was dismissed.


jcesarmobile:
I think, everything you put after [self.pickerController dismissViewControllerAnimated:NO completion:nil]; should be on the completion block instead

Anyway, there should be an issue created for this before it gets merged, open it as "New feature" on http://issues.cordova.io/

vladimir-kotikov:
@sarahgoldman, ping. Do you plan to finalize this? This might be a useful feature in some situations, but we need some steps to be completed first:

* As @jcesarmobile noticed, you need to create a JIRA for this feature. Also you'll probably need to get a community approval before proceed with this. You can send your proposal to dev@cordova.apache.org.

sarahgoldman:
I have created a Jira issue here: https://issues.apache.org/jira/browse/CB-9980

I have also emailed dev@cordova.apache.org with the proposal.

riknoll:
The dev list thread is [here](http://markmail.org/message/6jjxas3wrhwl6zb5). Looks like there is approval for this feature. @sarahgoldman I don't see your name in the [ICLA list](https://people.apache.org/committer-index.html#unlistedclas). You'll need to sign one (and potentially a CCLA) for this to be merged in. There are instructions for submitting an ICLA [here](http://cordova.apache.org/contribute/contribute_guidelines.html).

janpio:
Hey @sarahgoldman , there seems to be a merge conflict now. Could you maybe fix this? Thanks.

JBerendes:
Guessing she gave up on this. I usually give up on PRs within 6 months. 4 years of waiting doesn't inspire much confidence in a project/product.

jcesarmobile:
closing since the feature is iOS only and have not addressed the merge conflicts after a year

PR #118, opened 2015-08-27
Retaining image ratio when resizing

trubit:
According to documentation and iOS/Android behavior, resized images should retain ratio. This code fixes windows scaling.

### DIFF
--- a/src/windows/CameraProxy.js
+++ b/src/windows/CameraProxy.js
@@ -93,9 +93,9 @@ function resizeImage(successCallback, errorCallback, file, targetWidth, targetHe
             var image = new Image();
             image.src = imageData;
             image.onload = function() {
-                var scale = Math.min(targetWidth, targetHeight) / Math.max(this.width, this.height);
-                var imageWidth = scale * this.width;
-                var imageHeight = scale * this.height;
+                var ratio = Math.min(targetWidth / this.width, targetHeight / this.height);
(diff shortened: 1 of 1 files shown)


nikhilkh:
@muratsu Can you please review?

muratsu:
I think the logic needs some rework.
Imagine original picture with size 2w x 1h (aspect ratio 2). If we want to scale this image to 10x5 this CR will provide us with 5x2.5 which is not the most optimum outcome. 

Ideally the outcome should be:

original width x height | aspect ratio
--- | ---
2x1 | 2

 target width x height  | target aspect ratio | output width x height | output aspect ratio
--- | --- | --- | ---
10x5 | 2 | 10x5 | 2
12x5 | 2,4 | 10x5 | 2
8x5 | 1,6 | 8x4 | 2

trubit:
Yeah, in my case it worked, but correct calculation is:
    var scale = Math.min(targetWidth / this.width, targetHeight / this.height);

Calculates ratio for both sides and pick the small one

muratsu:
Yup, looks like right. Thanks!

vladimir-kotikov:
This was merged in https://github.com/apache/cordova-plugin-camera/commit/cad9ab0419ec3c90f0c052483177952d364a353d but not auto-closed. @trubit, close please.

vladimir-kotikov:
Thanks

PR #119, opened 2015-09-08
CB-9622 WP8 Camera Option destinationType:NATIVE_URI is a NO-OP

sgrebnov:
•	Windows Phone 8 proxy now supports NATIVE_URI param
•	Treat NATIVE_URI same way as FILE_URI (similar to Android)


vladimir-kotikov:
LGTM

PR #120, opened 2015-09-08
CB-9623 Fixes various issues when encodingType set to 'png'

vladimir-kotikov:
See bug description in [CB-9623](https://issues.apache.org/jira/browse/CB-9623)


nikhilkh:
LGTM

PR #122, opened 2015-09-09
CB-9633 iOS Taking a Picture With Option destinationType:NATIVE_URI not working

sgrebnov:
In case sourceType == Camera we receive nil as Uri because image is stored in memory. This PR adds logic to save image locally before obtaining resultant uri in this case


vladimir-kotikov:
LGTM. Rebasing and merging.

PR #124, opened 2015-09-10
[android] fix crop from gallery doesn't work 

EricCheung3:
fix crop from gallery doesn't work


vladimir-kotikov:
@EricCheung3, is this still actual? Are you sure this is correct?

I believe the crop should work fine with both PHOTOLIBRARY and CAMERA if your photo app have crop support. Please see https://github.com/apache/cordova-plugin-camera#android-quirks-1 for more details

infil00p:
Works for me, closing this really old issue

PR #127, opened 2015-09-30
update video and button style to display them on side-menu layout in ionic app

raykin:


riknoll:
I'm not familiar with the ionic side menu layout, but I think that #138 might be a more general solution. @raykin does that sound alright?

raykin:
yes, you're right, adding id is more scalable solution.

PR #128, opened 2015-10-01
Marshmallow/Cordova Android 5.0.x compatibility changes 

infil00p:
This patch also fixes a long-standing issue with files not actually being returned, and instead Content URIs being returned instead.  Since the CordovaResourceApi should already be able to handle this, we use it to find the file.  However, this doesn't work with files in the cloud that may not be in the device.


infil00p:
Closing pull request.  Code has now been rebased and merged.

PR #130, opened 2015-10-29
CB-7668 Adding a sterner warning for allowedit on Android

riknoll:
Adds a much clearer warning about `allowEdit` on Android to the README. That particular feature has been the cause of many JIRA issues (see CB-7668)


riknoll:
@infil00p please review

PR #132, opened 2015-11-03
CB-9910 android: Add permission request for some gallery requests

riknoll:
Fixes an issue where there were some permission requests missing when getting images from the PHOTOLIBRARY source. Any camera options that required outputting a modified image (width/height, allowEdit, correctOrientation, etc.) need the permission in order to preserve filenames and EXIF data in the final image.


riknoll:
@infil00p does this new permission request seem alright to you? We can remove it, but we will lose the EXIF data and filenames of images taken from the gallery.

infil00p:
That permission request looks fine.  I really wish we could communicate why we need the damn permission.

riknoll:
Yeah, I can add something to the quirks section of the README so that developers know what's going on, but that doesn't really help the end user.

PR #135, opened 2015-11-06
CB-9960 android: Fixed FileNotFoundException for some gallery images

riknoll:
Certain gallery applications (such as Google Photos) return `content://` URIs that our method of getting a file path (`FileHelper.getRealPath()`) fails to properly dereference. Using these applications with `cameraOptions` that cause the image to be edited (width/height, correctOrientation, allowEdit, etc.) would trigger a FileNotFoundException and cause the plugin to fail because it is unable to grab the original filename. This checks the result of `FileHelper.getRealPath()` and if it fails to grab a path instead uses the generic filename "modified.jpg" for the local file (which was the intended behavior before, we were just checking the result of `getRealPath()` incorrectly). Note: These images will lose their EXIF data (the Exifinterface needs a local file path, I think)

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -547,7 +547,7 @@ private String ouputModifiedBitmap(Bitmap bitmap, Uri uri) throws IOException {
         String realPath = FileHelper.getRealPath(uri, this.cordova);
 
         // Get filename from uri
-        String fileName = realPath != null ?
+        String fileName = realPath != null && !realPath.isEmpty() ?
             realPath.substring(realPath.lastIndexOf('/') + 1) :
             "modified." + (this.encodingType == JPEG ? "jpg" : "png");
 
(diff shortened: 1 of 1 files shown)


src/android/CameraLauncher.java
```diff
@@ -547,7 +547,7 @@ private String ouputModifiedBitmap(Bitmap bitmap, Uri uri) throws IOException {
         String realPath = FileHelper.getRealPath(uri, this.cordova);
 
         // Get filename from uri
-        String fileName = realPath != null ?
+        String fileName = realPath != null && !realPath.isEmpty() ?
```
jasongin: This is fine, but you may also consider [TextUtils.isEmpty(realPath)] (http://developer.android.com/reference/android/text/TextUtils.html#isEmpty(java.lang.CharSequence)) -- that is the recommended way in Android to check if a string is null or empty.

jasongin: Looks like TextUtils.isEmpty is not currently used anywhere in Cordova Android code, so I'm OK with leaving this as-is.

jasongin:
LGTM

infil00p:
Looks fine, but maybe we should just remove this check and do this another way.

riknoll:
I was trying to preserve the existing behavior, but I totally agree it's not ideal. What we lose by doing this check is the filename and EXIF data for images not on the device. I wonder if the filename is in fact retrievable using the ContentResolver (I haven't looked into it that closely). We might be able to grab the EXIF when we get the filestream in `getScaledbitmap()`. That might be a bit more robust overall.

infil00p:
I think that's what #112 does.

riknoll:
Seems to have been fixed by #175, so I'm closing this

PR #136, opened 2015-11-10
CB-9169 android: Fixed filetype for uncompressed images and added quirk

riknoll:
When trying to get uncompressed images on Android using camera options like these:
``` javascript
{
    destinationType : Camera.DestinationType.FILE_URL,
    sourceType : Camera.PictureSourceType.CAMERA,
    encodingType: Camera.EncodingType.PNG,
    quality: 100,
    saveToPhotoAlbum: true
}
```
the plugin will return a JPEG file with a .png filename. This pull request fixes the filename and documents the Android quirk that `encodingType` is ignored by the plugin on Android for uncompressed files. It currently does not make sense to actually provide PNG support in this case, because native camera apps return JPEG files and so converting to PNG does not improve quality, it just hurts performance. Same can be said for the gallery which just returns the selected file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -408,7 +408,7 @@ Can only return photos as base64-encoded image.
 
 #### Firefox OS Quirks
 
-Camera plugin is currently implemented using [Web Activities](https://hacks.mozilla.org/2013/01/introducing-web-activities/). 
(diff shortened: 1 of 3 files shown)


riknoll:
This doesn't fix CB-9169, but it documents the quirk on Android. @infil00p please review.

infil00p:
Looks good.

PR #137, opened 2015-11-11
CB-5479 android: changed saveToPhotoAlbum to save uncompressed images

riknoll:
It was apparently decided a while back (CB-5749) that we should be saving the uncompressed image with `saveToPhotoAlbum` rather than the modified image when the camera options `quality`, `targetHeight`, `targetWidth`, etc. are given. iOS supports this behavior, but Android does not. This pull request changes the Android code so that when `saveToPhotoAlbum` is specified, the uncompressed image is saved to the album and the modified image is saved to the application's cache (and that URI is what is returned to the application). This also makes it so that DATA_URL actually saves images to the gallery when the flag is enabled (this was previously not the case). This might be a breaking change for some applications that relied on the old behavior (saving one image rather than two).

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -441,7 +441,22 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
         }
 
         Bitmap bitmap = null;
-        Uri uri = null;
+        Uri galleryUri = null;
+
+        // CB-5479 When this option is given the unchanged image should be saved
+        // in the gallery and the modified image is saved in the temporary
+        // directory
+        if (this.saveToPhotoAlbum) {
+            galleryUri = Uri.fromFile(new File(getPicutresPath()));
+
+            if(this.allowEdit && this.croppedUri != null) {
+                writeUncompressedImage(this.croppedUri, galleryUri);
+            } else {
+                writeUncompressedImage(this.imageUri, galleryUri);
+            }
+
+            refreshGallery(galleryUri);
+        }
 
         // If sending base64 image back
         if (destType == DATA_URL) {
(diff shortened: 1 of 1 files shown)


riknoll:
@infil00p This might be breaking (sorta). What is the protocol we usually follow for merging breaking changes in?

riknoll:
Oh, [according to this](http://callback.markmail.org/search/?q=+saveToPhotoAlbum+list%3Aorg.apache.incubator.callback-dev#query:saveToPhotoAlbum%20list%3Aorg.apache.incubator.callback-dev%20date%3A201312%20+page:1+mid:v6gqvpfabvtja6lu+state:results) we should save the cropped image if `allowEdit` is on. I'll update the PR.

riknoll:
The commit I just pushed handles cropped images. I believe it also fixes an undocumented bug where requesting uncompressed/unedited images caused the app to ignore the crop Activity's output. That is, the user would get to crop the image and then the plugin would ignore that and just return the uncropped file.

riknoll:
I think this should also fix CB-6640

StopSine:
+1

riknoll:
@infil00p, if you have no objections, I'm going to merge this in. I don't think it'll break too many people and it fixes several quirks.

infil00p:
Sounds good.

luofan189:
Is this issue only fixed on Android side? I am seeing the same issue on iOS with the latest release as well.

riknoll:
@luofan189 What issue are you seeing? I can take a look.

luofan189:
@riknoll only the low resolution image is saved on iOS device as well. I set the targetWidth and targetHeight to 1024x768

riknoll:
@luofan189 I'll try and reproduce this. If you don't mind, let's move this conversation to the JIRA issue:

https://issues.apache.org/jira/browse/CB-5479

riknoll:
@luofan189 I would also appreciate it if you could post the camera options you are passing to the plugin and the versions of iOS and the plugin you're using. Thanks!

luofan189:
I have added options to that JIRA issue, @riknoll

PR #140, opened 2015-11-18
CB-8940 Setting z-index values to maximum for UI buttons

rakatyal:


purplecabbage:
Why 2147483646 ?

rakatyal:
I believe that is to keep the video element behind the buttons.

dblotsky:
You could place the value into a constant named `HIGHEST_POSSIBLE_Z_INDEX`, and then it would be both documented and easier to maintain.

rakatyal:
Good suggestion.

riknoll:
Agree with above. Otherwise, LGTM. Tested on an 8.1 device (and by tested I mean I took a picture, not really sure what else to do here). I don't really see how this can mess anything up.

PR #142, opened 2015-12-01
Fix missing CAMERA permission for Android M

ochakov:
Fixing a bug where for Android M new permissions model, only external storage permission was asked, while camera permission was not, causing camera to fail on Android 6.0

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -238,10 +238,10 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
      * @param returnType        Set the type of image to return.
      */
     public void callTakePicture(int returnType, int encodingType) {
-        if (cordova.hasPermission(permissions[TAKE_PIC_SEC])) {
+        if (cordova.hasPermission(permissions[TAKE_PIC_SEC]) && cordova.hasPermission(permissions[SAVE_TO_ALBUM_SEC])) {
             takePicture(returnType, encodingType);
         } else {
-            getReadPermission(TAKE_PIC_SEC);
+            cordova.requestPermissions(this, TAKE_PIC_SEC, permissions);
         }
     }
 
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -105,7 +105,7 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
     private boolean orientationCorrected;   // Has the picture's orientation been corrected
     private boolean allowEdit;              // Should we allow the user to crop the image.
 
-    protected final static String[] permissions = { Manifest.permission.READ_EXTERNAL_STORAGE };
+    protected final static String[] permissions = { Manifest.permission.CAMERA, Manifest.permission.READ_EXTERNAL_STORAGE };
 
     public CallbackContext callbackContext;
     private int numPics;
@@ -117,7 +117,7 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
 
     protected void getReadPermission(int requestCode)
     {
-        cordova.requestPermission(this, requestCode, Manifest.permission.READ_EXTERNAL_STORAGE);
+        cordova.requestPermission(this, requestCode, permissions[requestCode]);
     }
 
     /**
@@ -178,7 +178,7 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
                     // preserve the original exif data and filename in the modified file that is
                     // created
                     if(this.mediaType == PICTURE && (this.destType == FILE_URI || this.destType == NATIVE_URI)
-                            && fileWillBeModified() && !cordova.hasPermission(permissions[0])) {
+                            && fileWillBeModified() && !cordova.hasPermission(permissions[SAVE_TO_ALBUM_SEC])) {
                         getReadPermission(SAVE_TO_ALBUM_SEC);
                     } else {
                         this.getImage(this.srcType, destType, encodingType);
@@ -238,7 +238,7 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
      * @param returnType        Set the type of image to return.
      */
     public void callTakePicture(int returnType, int encodingType) {
-        if (cordova.hasPermission(permissions[0])) {
+        if (cordova.hasPermission(permissions[TAKE_PIC_SEC])) {
             takePicture(returnType, encodingType);
         } else {
             getReadPermission(TAKE_PIC_SEC);
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -58,6 +58,9 @@ import android.provider.MediaStore;
 import android.util.Base64;
 import android.util.Log;
 import android.content.pm.PackageManager;
+import android.content.pm.PackageManager.NameNotFoundException;
+import android.content.pm.PermissionInfo;
+
 /**
  * This class launches the camera view, allows the user to take a picture, closes the camera view,
  * and returns the captured image.  When the camera view is closed, the screen displayed before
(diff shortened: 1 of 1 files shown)


riknoll:
Thanks for submitting a PR! I actually just ran into this issue while working on [another issue in JIRA](https://issues.apache.org/jira/browse/CB-10064). After doing some research into this problem, I found that it actually only comes up if you have the CAMERA permission declared in your manifest (I have no idea what the reasoning behind this behavior is). We don't declare that permission, but third party plugins can insert entries into the manifest and so we still need to handle the case. Anyway, the question is, should we always request the camera permission or should we check the manifest for the permission declaration first? I'm leaning towards checking the manifest (we shouldn't get extra permissions for no reason). @infil00p thoughts on that?

riknoll:
BTW, I created a JIRA for this ([CB-10120](https://issues.apache.org/jira/browse/CB-10120))

ochakov:
I have two questions then:
1. When you do not declare camera permission in the manifest, does it work on older Android versions?
2. When you do not declare camera permission, what is the result of cordova.hasPermission(CAMERA)? We can tweak the function to request only those permissions for which cordova.hasPermission returns false.

infil00p:
You do not need the Camera permission for the Camera plugin because we use intents.  Why are we trying to request this permission?  That's the recommended way of using the Camera as per the Android Guidelines, and it's so you don't have to have the permission in the manifest.

I'm going to close this issue and I don't think we should accept this request, since we're requesting a permission that we don't actually need.

riknoll:
@infil00p if a third party plugin declares the permission in their plugin.xml so that is inserted into the manifest, then you get a SecurityException and a crash. I have reproduced that behavior.

riknoll:
In the developer reference: http://developer.android.com/reference/android/content/pm/PackageInfo.html#requestedPermissions

infil00p:
@riknoll Yeah, but that's a problem that should be in the docs, and not in this code.

infil00p:
@ochakov 

1. Yes, it still works, because this relies on the permission of the Camera, not the permission of the application firing the event.
2. If you don't declare camera permission, the result of cordova.hasPermission(CAMERA), assuming CAMERA is the string containing the Camera permission, would be false.  We shouldn't request permissions for things that we're not actually doing, such as taking a photo.

riknoll:
I feel like we can easily handle it, though, by programmatically checking if the permission was declared before we take a picture and only requesting it if it is (looks like it's in the [PackageInfo](http://developer.android.com/reference/android/content/pm/PackageInfo.html#requestedPermissions)). Plus, we avoid another quirk for the Android platform. Is there any danger in requesting that permission if some other plugin is already using it?

infil00p:
BTW: Do we have a third-party plugin that asks for Camera? And if a Third-Party plugin is asking for the CAMERA permission, shouldn't they have their own activity that actually uses the camera? This seems like a weird hypothetical situation.

riknoll:
Here's one that declares it: https://github.com/phonegap/phonegap-plugin-barcodescanner

infil00p:
Right, Barcode Scanner! That thing! So, if there's an app that uses the Barcode Scanner and takes a picture, then it can cause a problem. 

I'm still not stoked about this, but this code should probably be added.

riknoll:
Agreed. But this PR should be updated to check the PackageInfo first. @ochakov does that sound alright to you?

ochakov:
riknoll,
Sounds alright. I would also check camera and storage permissions separately in case someone will approve one of them, but deny the other. Next time, only the missing one should be asked. I will commit another update for that.

ochakov:
Added both fixes to this PR
Camera permission will not be required unless android.permission.CAMERA is set in the package.
BTW, barcode scanner and flashlight plugins both set that permission in the manifest.

I tested with and without android.permission.CAMERA and it works well in both cases, not asking for camera permission in case when both plugins are uninstalled.

riknoll:
@ochakov I'll do some testing, but this looks pretty good! In the meantime, can you fill out an Apache ICLA? There are instructions for doing so [here](http://cordova.apache.org/contribute/contribute_guidelines.html). That needs to be submitted/filed before I can merge this in.

src/android/CameraLauncher.java
```diff
@@ -178,7 +181,7 @@ else if ((this.srcType == PHOTOLIBRARY) || (this.srcType == SAVEDPHOTOALBUM)) {
                     // preserve the original exif data and filename in the modified file that is
                     // created
                     if(this.mediaType == PICTURE && (this.destType == FILE_URI || this.destType == NATIVE_URI)
-                            && fileWillBeModified() && !cordova.hasPermission(permissions[0])) {
+                            && fileWillBeModified() && !cordova.hasPermission(permissions[SAVE_TO_ALBUM_SEC])) {
```
riknoll: I don't think we should be using SAVE_TO_ALBUM_SEC and TAKE_PICTURE_SEC for the indices in the permissions array since those values mean something different. The action after the permissions is not determined by the permissions requested (both the camera and the photolibrary sources require READ permission, for example). Could you define some different constants at the top of the file; maybe CAMERA_PERMISSION_INDEX and READ_PERMISSION_INDEX?

ochakov:
ICLA has been filed and approved by Craig L Russell

riknoll:
@ochakov thanks for doing that! Can you add your name to your GitHub profile so that we can verify?

riknoll:
@ochakov any updates on this?

ochakov:
@riknoll done

riknoll:
Awesome, thanks for doing that! Would you mind also incorporating the fix I commented on in the diff?

ochakov:
@riknoll, I can't really invest more time in this bug. I am doing completely different things now.

ianisted:
Any luck with a resolution to this issue? I'm using that Barcode plugin and I'm getting the following error "Sorry, the Android camera encountered a problem...". After much head scratching I discovered it didn't have permissions to access the Camera, and wasn't asking for them.

ochakov:
@ianisted 
Use my version of the plugin
https://github.com/ochakov/cordova-plugin-camera

ianisted:
@ochakov thanks, but I'm needing to perform QR code scanning. Seems you have a similar issue, though the phonegap-plugin-barcodescanner plugin uses Intent and not Permissions.

MatthewBooth:
This should be merged ASAP. It fixes this very issue for a project I'm working on. 

You absolutely should be declaring the permissions in the manifest so that the user can see them on the Google Play Store, or while installing manually. It's poor practice to deny the user the chance to remove, revoke or review permissions.

I'm using @ochakov's fork currently, but it would obviously be better if this was in the official channels.

infil00p:
WTF? I thought this was merged weeks ago, or we decided not to merge it or something.  We need to get the conflicts resolved and this thing pulled in.

MatthewBooth:
@infil00p I've rebased here: https://github.com/apache/cordova-plugin-camera/pull/174

dieppe:
You can also check this fork https://github.com/jnuine/cordova-plugin-camera

It includes a major refactor of the iOS plugin which fixes some bugs, checks for option consistency and allows access to the picture metadata in all cases that make sense (available as a [PR](https://github.com/apache/cordova-plugin-camera/pull/101) since June last year, but I guess merging PR is not cordova's team strong suit).

On Android, it includes these commits and also handles better the kitkat+ gallery (sadly not the images from Google Drive).

We use it in an app used daily both on Android and iOS.

okonon:
Hello all! @ochakov could you please resolve the conflicts so this can be merged?
Or anyone else?
@riknoll @infil00p @MatthewBooth 

As more people are upgrading to Android M this becomes more common issue out in the while.
Thanks

riknoll:
Hey @okonon, that has been done in #179, which will hopefully be merged in soon. Just waiting on an ICLA.

SergiiS:
Hello guys, was it merged? Thanks!

mccob:
how to invoke permission since Android M permissions system for WebRTC getUserMedia ?

PR #144, opened 2015-12-03
[iOS] Fixed authorization error when user clicks cancel in image selector view.

nleclerc:
I'm resurrecting a change from [an old pull request](https://github.com/apache/cordova-plugin-camera/pull/24) that was closed without any explanation.

[The original issue](https://issues.apache.org/jira/browse/CB-6576) has been improperly marked as resolved when it clearly is not.

Problem is there's currently no way to know when a user has just cancelled the image capture since clicking cancel send a `has no access to assets` error when it really isn't an authorization issue.

### DIFF
--- a/src/ios/CDVCamera.m
+++ b/src/ios/CDVCamera.m
@@ -518,7 +518,7 @@ static NSString* toBase64(NSData* data) {
     
     dispatch_block_t invoke = ^ (void) {
         CDVPluginResult* result;
-        if ([ALAssetsLibrary authorizationStatus] == ALAuthorizationStatusAuthorized) {
+        if (picker.sourceType == UIImagePickerControllerSourceTypeCamera || [ALAssetsLibrary authorizationStatus] == ALAuthorizationStatusAuthorized) {
             result = [CDVPluginResult resultWithStatus:CDVCommandStatus_ERROR messageAsString:@"no image selected"];
         } else {
             result = [CDVPluginResult resultWithStatus:CDVCommandStatus_ERROR messageAsString:@"has no access to assets"];


riknoll:
@vladimir-kotikov can you check this one out? Looks simple enough, but iOS is not my forte.

riknoll:
Also, the JIRA is here: [CB-10128](https://issues.apache.org/jira/browse/CB-10128)

m-revetria:
I don't think this change completely fix the issue: we should check AVCaptureDevice auth status just when the picker source is the camera and check ALAssetsLibrary auth status just when the picker source type is photo library or saved photos album.

Suppose next case:
1. open the image picker with source type camera
2. the user doesn't grant access to the device's camera
3. finally closes the UIImagePickerController by pressing the "Cancel" button. 

In that case the error message "no image selected" is returned, I would spect a message like "no access to camera".

> Notice that this could happen just the first time the image picker is shown with source type camera, plugin is checking auth status before presenting the image picker.

You can take a look to full auth check in my commit https://github.com/m-revetria/cordova-plugin-camera/commit/bbdafb08640cafc8e98ed0cbd5d558ad9b6dc150

vladimir-kotikov:
After couple of tests this giving me a "no access to assets" error when cancelling a camera. @m-revetria solution looks more correct and robust.

nleclerc:
Well, i'll gladly go with Miguel's solution.
What should be the next step ?

riknoll:
Hey @m-revetria, would you mind making a PR for that change?

m-revetria:
Hi, 
@riknoll: I just made a PR for this change: https://github.com/apache/cordova-plugin-camera/pull/146:

riknoll:
Hey @nleclerc, mind closing this PR? #146 was merged

nleclerc:
ok.

PR #145, opened 2015-12-04
CB-9189 android: Implementing save/restore API to handle Activity destruction

riknoll:
This fixes the long standing issue of the camera plugin crashing after receiving an Activity result when the Cordova Activity is killed in the background (CB-9189). Uses the new API for plugin save/restore that was merged into cordova-android recently.

*NOTE*: The link I put in the README is for Android docs that have been merged in to dev, but not published yet. It'll work once we do a docs release. Right now it's broken.

### DIFF
--- a/README.md
+++ b/README.md
@@ -400,7 +400,12 @@ scenario, the image may not appear when the cordova activity is restored.
 
 Android uses intents to launch the camera activity on the device to capture
 images, and on phones with low memory, the Cordova activity may be killed.  In this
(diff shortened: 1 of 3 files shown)


riknoll:
Rebased to master and update the README link. Sorry, I lost track of this over the holidays. I plan to merge this in at end of day. @jasongin can you give this a quick look?

jasongin:
LGTM

PR #148, opened 2015-12-11
fix Android 4.4 content:// URI transform to file:/// URI error

gaoyuan123:
2 bug fixed:
1.android 4.4 FileHelper.getRealPath will return null lead to can not create the picture
2.android 4.2 FileHelper.getRealPathFromURI_API11to18 return null lead to can not create the picture

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -18,31 +18,12 @@
 */
 package org.apache.cordova.camera;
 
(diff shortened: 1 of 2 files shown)


riknoll:
Thanks for the PR! Can you post some Javascript so that I can try and reproduce these issues and test your fix? Also, I can't accept this PR with all of the changes to the import statements; Could you rebase so that those aren't present?

PR #153, opened 2016-01-16
CB-10370: Added option useGeolocation. Also fixed saving geolocation.

vldmrrr:
While adding the option I found that saving geolocation was broken - result was returned to javascript before location manager code had a chance to update image data. The fix for that is included. I am not sure if a separate CB needed for that.


nikhilkh:
This code has some merge conflicts. It makes sense to move away from a preference and use an option. Also, now the docs for plugin-camera are generated using jsdoc comments. It's best to close this PR and re-create a new one.

infil00p:
This code still has merge conflicts, so it's getting closed.

PR #155, opened 2016-01-20
Fix FileNotFoundException in ouputModifiedBitmap

MiniGab:
The ouputModifiedBitmap expect the result from FileHelper.getRealPath to be null in case of error :

    // if realPath == "" => fileName = ""

    String modifiedPath = getTempDirectoryPath() + "/" + fileName;

    OutputStream os = new FileOutputStream(modifiedPath); <-- error when fileName == ""

This might be related to https://issues.apache.org/jira/browse/CB-9960 and https://issues.apache.org/jira/browse/CB-10093


riknoll:
See #135

PR #158, opened 2016-01-21
[CB-10388] - Added correctOrientation support for android 4.4

horozco:
Jira issue: https://issues.apache.org/jira/browse/CB-10388
- Currently If You pass the correctOrientation param as true, the sourceType as PHOTOLIBRARY or SAVEDPHOTOALBUM, the media type as ALLMEDIA and the selected file is an image the correctOrientation function is not being called. Also even if the function is called the code is not detecting the URI and doesn’t rotate the picture. These changes fixed both issues.

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -623,9 +623,11 @@ private String ouputModifiedBitmap(Bitmap bitmap, Uri uri) throws IOException {
         String fileLocation = FileHelper.getRealPath(uri, this.cordova);
         Log.d(LOG_TAG, "File locaton is: " + fileLocation);
 
-        // If you ask for video or all media type you will automatically get back a file URI
+        // If you ask for video type, or you ask for all media type and the selected type is not an image
+        // you will automatically get back a file location
         // and there will be no attempt to resize any returned data
-        if (this.mediaType != PICTURE) {
+        String type = cordova.getActivity().getContentResolver().getType(uri);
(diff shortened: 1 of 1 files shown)


riknoll:
Thanks for submitting a PR! The ALLMEDIA thing definitely sounds like a bug and your change there looks good.

However, I'm not 100% sure about the changes you made to `getImageOrientation()`. The issue is that when we get an image URI, it may or may not map to an actual file location depending on the application that provided the URI. For example, Google Photos can provide URIs that point to images in the cloud and not on the physical device. In this case, using the content resolver is the correct method for getting the orientation because the content resolver knows how to handle those situations. Thoughts?

fmiboy:
Hello guys,
Currently I have image orientation correct when I use PNG type and on Android 4.4.4 but Android 5 orientation doesn't work at all.
Any idea when there will be fix?

Thanks

morioma:
I prefer this one to https://github.com/apache/cordova-plugin-camera/pull/197, because that one has code duplication for calculating rotation.
I am thinking getting rid of the ```getImageOrientation``` function all together and use the ```getRotatedBitmap``` from ```processResultFromCamera``` which based on the EXIF helper.

Additionally, the issue may not just for android 4.4. Latest Android 6.0.1 on a Galaxy S6 edge suffers the same issue.

swbradshaw:
@morioma  - Take a look at #197 again. I completely rewrote the implementation.

infil00p:
Closing this since #197 got merged in

PR #160, opened 2016-02-01
CB-10270 android: Added back support for file:// URIs to getRealPath

riknoll:
Our `getRealPath()` function is unable to handle `file: //` URIs, which can occasionally lead it to return an empty string for the result URI. These camera options reproduce when you select an image from the Gallery app:
```
{
    destinationType: Camera.DestinationType.FILE_URI,
    sourceType: Camera.PictureSourceType.PHOTOLIBRARY,
    allowEdit: true,
    correctOrientation: true
}
```
I believe that we used to do something similar to this, but it was removed at some point in the past. @infil00p do you have any idea why that was?

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -688,11 +688,17 @@ private String ouputModifiedBitmap(Bitmap bitmap, Uri uri) throws IOException {
 
                         } catch (Exception e) {
                             e.printStackTrace();
-                            this.failPicture("Error retrieving image.");
+                            this.failPicture("Error modifying image");
                         }
                     }
                     else {
-                        this.callbackContext.success(fileLocation);
+                        if (fileLocation != null) {
+                            this.callbackContext.success("file://" + fileLocation);
+                        } else if (uriString != null) {
+                            this.callbackContext.success(uriString);
+                        } else {
+                            this.failPicture("Error retrieving image");
+                        }
                     }
                 }
                 if (bitmap != null) {
--- a/src/android/FileHelper.java
+++ b/src/android/FileHelper.java
@@ -50,7 +50,17 @@ public class FileHelper {
     public static String getRealPath(Uri uri, CordovaInterface cordova) {
         String realPath = null;
 
-        if (Build.VERSION.SDK_INT < 11)
+        if (uri.getScheme().equals("file")) {
+            String path = uri.getPath();
+            if (!path.startsWith("/android_asset/")) {
+                return path;
+            } else {
+                LOG.d(LOG_TAG, "Cannot get real path of an android_asset URI");
+                return null;
+            }
+        }
+
+        else if (Build.VERSION.SDK_INT < 11)
             realPath = FileHelper.getRealPathFromURI_BelowAPI11(cordova.getActivity(), uri);
 
         // SDK >= 11 && SDK < 19
@@ -209,7 +219,7 @@ public class FileHelper {
         }
         return MimeTypeMap.getSingleton().getMimeTypeFromExtension(extension);
     }
-    
+
     /**
      * Returns the mime type of the data specified by the given URI string.
      *


riknoll:
For reference, [this](https://github.com/apache/cordova-plugin-camera/commit/929733b8913311f2cfd504937268a8a1d22f4f9a) is the commit I mentioned.

shaoner:
@riknoll, somehow an exception is triggered by this line for some uri:

```java
try {
    String wholeID = DocumentsContract.getDocumentId(uri);
    // ...
} catch(Exception e) {
    filePath = "";
}
```

The exception looks actually like:

> java.lang.IllegalArgumentException: Invalid URI: content://media/external/images/media/5230

anuj6117:
I am also facing the same exception if I use ALLMEDIA.

riknoll:
@shaoner @anuj6117 Is that exception related to [the JIRA](https://issues.apache.org/jira/browse/CB-10270) this PR is addressing? That is, does it result in you getting an empty string back from the `getPicture()` call?

anuj6117:
Yes, I am getting an empty string back from the getPicture() call.

shaoner:
@riknoll yes, actually if you don't set correctOrientation to true, getRealPath is never called, it returns the uri instead (I don't see why) so this is not empty:
```java
             if (this.targetHeight == -1 && this.targetWidth == -1 &&
                     (destType == FILE_URI || destType == NATIVE_URI) && !this.correctOrientation) {
                 this.callbackContext.success(uri.toString());
```

However, I think we have two issues here:
1. getRealPath should be called even with correctOrientation set to false
2. `FileHelper.getRealPathFromURI_API19` throws an exception, returning an empty path in some cases:
   - happens with my galaxy S4 device (android v5.0.1) and I guess it cannot be reproduced with a S4 emulator unless the same rom is used 
   - does not happen with a nexus S emulator (android v5.1.1)

( this seems to be the case for LG G3 as well, http://stackoverflow.com/questions/2789276/android-get-real-path-by-uri-getpath/33074944#33074944 )

For now, I've got it working by returning `getRealPathFromURI_BelowAPI11` if the path is empty.

olexme:
I had the same issue (empty string from getPicture() when trying to get the gallery picture) and I've just tried to use this PR by
cordova plugin add https://github.com/MSOpenTech/cordova-plugin-camera#ee06d3efb233d5404d0aeda2abe08e6fd765c7fe
Unfortunatelly it didn’t solve the issue - instead the empty string I get “file://“.
My options:
options = {
    quality:            80,
    encodingType:       0,        
    destinationType:    1,       
    sourceType:         0,
    allowEdit:          false,
    saveToPhotoAlbum:   false,
    correctOrientation: true,
    cameraDirection:    1
};
Android 5.0.0 on Galaxy S5

riknoll:
Alright, the real solution to this problem in my opinion is to get rid of the `getRealPath` function altogether. Not all URIs have real paths and there are always going to be some cases that fail. The "Android" way of handling this is to use ContentResolvers in order to get input streams and to stop trying to force everything into being a file on the device. Unfortunately, that is a major rewrite of this plugin. The sheer number of JIRA issues and open PRs that deal with it should be justification enough, though.

As for this PR, it does patch over one issue and so I'm inclined to leave it open as is. The other issue you are seeing is related but requires a very different fix. I believe there is already a JIRA open tracking it but I can't recall the issue ID off the top of my head.

shahin8r:
This doesn't seem to fix empty string for videos (mediaType: 1) but is working fine on photos.

TanaseButcaru:
I can confirm this is working for picture selection.
@shahin8r you could use this commit to [prevent empty video content uri](https://github.com/TanaseButcaru/cordova-plugin-camera/commit/e032811c1cf2b382f8576d92cb04e41e0193e221) in case ``getRealPath`` fails to resolve it. You can test my ``unofficial`` branch.

The combination between this, my commit for video empty uri & the [cordova-plugin-filepath](https://github.com/hiddentao/cordova-plugin-filepath) plugin is the best workaround at the moment to get ``file://`` uri for selected media file.

riknoll:
Seems to have been fixed by #175, so I'm closing this

jcesarmobile:
#175 only fixes it on API 19 and greater, maybe it should be moved outside.

jcesarmobile:
I'm going to send a PR that will use the API 19 method on android 11 and greater and will handle file urls

PR #162, opened 2016-02-04
CB-10541: Changing default maxResoltion to be highestAvailable for Ca…

rakatyal:
…meraCaptureUI

### DIFF
--- a/src/windows/CameraProxy.js
+++ b/src/windows/CameraProxy.js
@@ -710,7 +710,10 @@ function takePictureFromCameraWindows(successCallback, errorCallback, args) {
     var UIMaxRes = WMCapture.CameraCaptureUIMaxPhotoResolution;
     var totalPixels = targetWidth * targetHeight;
 
-    if (totalPixels <= 320 * 240) {
+    if (targetWidth == -1 && targetHeight == -1) {
+        maxRes = UIMaxRes.highestAvailable;
+    }
+    else if (totalPixels <= 320 * 240) {
         maxRes = UIMaxRes.verySmallQvga;
     } else if (totalPixels <= 640 * 480) {
         maxRes = UIMaxRes.smallVga;


riknoll:
So does this still leave us open to the possibility of having the plugin crash when certain combinations of width and height are passed?

src/windows/CameraProxy.js
```diff
@@ -710,7 +710,10 @@ function takePictureFromCameraWindows(successCallback, errorCallback, args) {
     var UIMaxRes = WMCapture.CameraCaptureUIMaxPhotoResolution;
     var totalPixels = targetWidth * targetHeight;
 
-    if (totalPixels <= 320 * 240) {
+    if (targetWidth == -1 && targetHeight == -1) {
```
nikhilkh: Should we just initialize `maxRes` to higestAvailable on declaration instead of putting another if?

rakatyal: If targetWidth and targetHeight are not specified, they are initialized to -1. So even if we initialize, it will be changed to verySmallQvga in the present if condition.

rakatyal:
@riknoll: Yes. I have created a separate Jira bug to tackle that. We are waiting for the response from Windows team on the approach.

src/windows/CameraProxy.js
nikhilkh: Of course! Totally missed it.

riknoll:
Alright. LGTM

XiMnet:
Hi,


We are away for the Chinese New Year celebration from 8 February 2016 to 12 February 2016. We will be back on 15th February 2016 (Monday).


For any request which requires immediate attention, kindly contact Ms. Inez (012 &ndash; 3211 265) or Mr. Law (017 &ndash; 218 5918).


Thank you very much for your continuous support. We wish you a Happy Chinese New Year and happy holidays.


Cheers.

XiMnet Malaysia Sdn Bhd

PR #163, opened 2016-02-05
CB-10539: Commenting out the verySmallQvga maxResolution option

rakatyal:
This is a workaround for a bug on Windows side till it gets resolved.


nikhilkh:
LGTM.

PR #168, opened 2016-02-13
CB-10369: Added getVideo() feature with thumbnail creation (readded)

TanaseButcaru:
I've implemented the `getVideo` function for cordova-plugin-camera which gives the user the ability to record a video with default camera app.

You can create a sample cordova app and copy the www files from `samples/android/www_getVideo` and run an interactive test of this feature with all the options, as follows:
 *          `quality`: points to `Camera.VideoQuality`,
 *          `sourceType`: points to `Camera.VideoSourceType`,
 *          `encodingType`: points to `Camera.VideoEncodingType`,
 *          `mediaType`: points to `Camera.MediaType`,
 *          `mediaThumbnail`: points to `Camera.MediaThumbnail`,
 *          `mediaDurationLimit`: numeric value, 0 = disabled
 *          `mediaSizeLimit`: numeric value, 0 = disabled  

... where:
```js
VideoEncodingType:{
    /** Return 3GP encoded video */
    GPP: 2,
    /** Return MP4 encoded video */
    MP4: 3,
    /** Return WEBM encoded video */
    WEBM: 4,
    /** Return MKV encoded video */
    MKV: 5
  },

MediaThumbnail:{
    /** Media thumbnail disabled */
    NONE: 0,
    /** Creates a thumbnail of 512x384 px */
    MINI_KIND: 1,
    /** Creates a thumbnail of 96x96 px */
    MICRO_KIND: 2
 },

VideoSourceType:{
    /** Choose video from video library */
    VIDEOLIBRARY : 0,
    /** Take video from camera */
    CAMERA : 1
  },

VideoQuality:{
    /**  Low quality, suitable for MMS messages */
    LOW : 0,
    /** High quality */
    HIGH: 1
  },
```
I always thought that this plugin would launch the camera app in recording mode if `sourceType` set to CAMERA and `mediaType` set to VIDEO, but when I needed this I found that it's not implemented, so I made it myself.

I'm just a web dev, not a software developer, but I managed to get it done; this is my first contribution to a cordova plugin with native code, so there might be things that are not written well or missing checks. I'm open to suggestions and contributions for this.
Android only for the moment. If there's someone that can port this to iOS would be much appreciated, but in the meantime, I'll try my luck with that too.

Tested with two real devices: Sony Xperia Z3 Compact (5.1.1) and Sony Xperia Sola (unofficial 4.4.4)
Here are two screenshots with the interactive sample:


infil00p:
This is a new feature, and it technically doesn't belong in the Camera, which is designed to only take photos.

jcesarmobile:
@infil00p see comments on https://issues.apache.org/jira/browse/CB-10369, I commented that in the 2015 plugin audit there were recommendations to deprecate the media-capture plugin and use camera instead for the video, so this PR might be a good starting point. But maybe it should be discussed again on the mail list

janpio:
Hey @tbutcaru , there seems to be a merge conflict now. Could you maybe fix this? Thanks.

PR #169, opened 2016-02-15
separate FILE_URI and NATIVE_URI / take in account user defined destType

TanaseButcaru:
Right now, on Android, `FILE_URI` and `NATIVE_URI` returns the path with `content://` scheme, excepting the case when the image has been modified.

These changes take in account what is the user defined `destType` and returns the path as requested.

I know there were discussions on JIRA ([CB-9548](https://issues.apache.org/jira/browse/CB-9548)) about how `FILE_URI` and `NATIVE_URI` should look, but I came to think that it should be like this:
- FILE_URI should return the `file://` scheme;
- NATIVE_URI should return the `content://` scheme;

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -488,7 +488,10 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
                 // If we saved the uncompressed photo to the album, we can just
                 // return the URI we already created
                 if (this.saveToPhotoAlbum) {
-                    this.callbackContext.success(galleryUri.toString());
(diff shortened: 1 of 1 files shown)


riknoll:
I'm not sure we ever got consensus on the mailing list about what the desired behavior should be. I'm not saying this is wrong, I'm just not sure where we landed on this particular issue.

infil00p:
The web doesn't know what content URIs are, so we should never return them.

infil00p:
Closing PR, since it's conflicting and we're now using File Providers.

PR #149, opened 2015-12-12
Android FileHelper modifications to manage videos

dlogo:
Updated getRealPathFromUri method for APIs above 10th in order to manage not only images but also videos. Furthermore the function for API11to18 crashed if uri format is like this: 'content://media/external/images/media/350259'.

PR #170, opened 2016-02-15
Fix FileHelper / port cordova-plugin-filepath-unofficial and replace getRealPath with it

TanaseButcaru:
I recently encountered the `FileNotFoundException ouputModifiedBitmap` error while using this plugin and I've tried all PRs available here but none seems to fix the issue.
The current implementation of FileHelper does not deal well with files from external (and even internal) SD card (removable or not).
I also had issues getting videos file paths.

I was using [cordova-plugin-filepath](https://github.com/TanaseButcaru/cordova-plugin-filepath-unofficial) to get the `file://` path from a `content://` one. 
This plugin had its one issues, which I managed to fix & improve it as a whole. 

Because the filepath plugin is doing its job well, I ported it to the cordova-plugin-camera. This looks a lot like PR #93 , just that it is updated.
The only case where this fails (as far as I have tested) is the case when user selects a picture/video from Google Drive (there's no solution for this in #143 PR too).

This closes the following:
- #160 
- #155 
- #149
- #148 
- #143 
- #135 
- #112 
- #93  

You can test this PR & the #169 PR from my [unofficial version release](https://github.com/TanaseButcaru/cordova-plugin-camera-unofficial) of cordova-plugin-camera.

### DIFF
--- a/src/android/FileHelper.java
+++ b/src/android/FileHelper.java
@@ -25,6 +25,9 @@ import android.os.Build;
 import android.provider.DocumentsContract;
 import android.provider.MediaStore;
 import android.webkit.MimeTypeMap;
+import android.content.ContentUris;
+import android.util.Log;
+import android.os.Environment;
 
 import org.apache.cordova.CordovaInterface;
 import org.apache.cordova.LOG;
@@ -33,6 +36,8 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Locale;
+import java.util.List;
+import java.io.File;
 
 public class FileHelper {
     private static final String LOG_TAG = "FileUtils";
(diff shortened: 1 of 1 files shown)


davidbarre:
any update on this ?

davidbarre:
#160 #155 #149 #148 #93 are close
#143 #112 are still opened
So what is the status of this PR ? 

A bag of thing are in master branch ? is next release planified ?

swbradshaw:
Someone needs to take a look at all the open camera PR.  There are many problems with this getRealPath and it has been discussed in a few places. It seems like this PR helps out a lot. For what its worth I removed the need for this function in  #197 for most cases.  (#197 also allows user to pick a photo from Google Drive.)

davidbarre:
Hi @TanaseButcaru,

Do you continue to upgrade your unoffical plugin with version 2.3 ?
Or do you consider that version 2.3 of official plugin fix this issues ?

Thanks

--
David

TanaseButcaru:
Hi @davidbarre,

At the moment, I don't have the time to continue upgrading my unofficial plugin with version 2.3.
The last time I have made changes to my unofficial plugin, the official implementation wasn't covering all the cases.

I may try to check the differences between my unofficial version and the latest official in the next days.
I'll keep up updated.

infil00p:
There's tons of problems with getRealPath, we should be using the FilesProvider to access files instead, since it actually maps to real locations on the disk instead of messing with the MediaStore.

PR #172, opened 2016-02-18
CB-10552 Replacing images in README.md.

dblotsky:


dblotsky:
@riknoll @nikhilkh @sarangan12 please review when you have a moment.

daserge:
@dblotsky, should not these changes be made in js source instead?
Then you need to re-generate the docs using [`npm install && npm run gen-docs`](https://github.com/apache/cordova-plugin-camera/tree/61b77951e11cfdae8eaf5e6b2f8392b4822239f5#this-documentation-is-generated-by-a-tool).

dblotsky:
@daserge thanks for pointing that out. I made the change in the source. However, this was not a very intuitive process. Can this be done automatically?

daserge:
@dblotsky, docs generation will run automatically as a pre-commit hook but you need to do `npm install` first. I agree this is not very intuitive though, and the corresponding dicsussion is [here](https://github.com/cordova/cordova-discuss/pull/16) in case you have some thoughts on improving the process.

daserge:
We could also add a Travis build step verifying that the `npm run gen-docs > TEMP_README.md` equals `README.md`.

dblotsky:
@daserge Thanks for pointing me to the discussion. I forgot about it! I think right now the process is unintuitive. Now that the plugin docs are moving to the main site, perhaps we can rethink this.

PR #175, opened 2016-02-21
Fix for CB-10625

jcesarmobile:
getDocumentId was crashing in some cases.
Now, in case it crashes, it will use the original uri to query.


infil00p:
OK, this breaking the CI could mean anything.

PR #177, opened 2016-02-22
CB-10639 Appium tests: Added some timeouts,

alsorokin:
Taking a screenshot on failure,
Retry taking a picture up to 3 times,
Try to restart the Appium session if it's lost

https://issues.apache.org/jira/browse/CB-10639

### DIFF
--- a/appium-tests/android/android.spec.js
+++ b/appium-tests/android/android.spec.js
@@ -1,36 +1,43 @@
-/*jslint node: true, plusplus: true */
-/*global beforeEach, afterEach */
-/*global describe, it, xit, expect, jasmine */
-'use strict';
+/*jshint node: true, jasmine: true */
 
 // these tests are meant to be executed by Cordova Medic Appium runner
 // you can find it here: https://github.com/apache/cordova-medic/
 // it is not necessary to do a full CI setup to run these tests
 // just run "node cordova-medic/medic/medic.js appium --platform android --plugins cordova-plugin-camera"
 
+'use strict';
+
 var wdHelper = require('../helpers/wdHelper');
 var wd = wdHelper.getWD();
 var cameraConstants = require('../../www/CameraConstants');
 var cameraHelper = require('../helpers/cameraHelper');
+var screenshotHelper = require('../helpers/screenshotHelper');
+
+var STARTING_MESSAGE = 'Ready for action!';
+var RETRY_COUNT = 3; // how many times to retry taking a picture before failing
+var MINUTE = 60 * 1000;
+var DEFAULT_SCREEN_WIDTH = 360;
+var DEFAULT_SCREEN_HEIGHT = 567;
+var DEFAULT_WEBVIEW_CONTEXT = 'WEBVIEW';
 
 describe('Camera tests Android.', function () {
-    var driver,
-        startingMessage = 'Ready for action!',
-        // the name of webview context, it will be changed to match needed context if there are named ones:
-        webviewContext = 'WEBVIEW',
-        // this indicates if device library has test picture:
-        isTestPictureSaved = false,
-        // this indecates that there was critical error and tests cannot continue:
-        stopFlag = false,
-        // we need to know the screen width and height to properly click on the first image in the gallery
-        screenWidth = 360,
-        screenHeight = 567;
+    var driver;
+    // the name of webview context, it will be changed to match needed context if there are named ones:
+    var webviewContext = DEFAULT_WEBVIEW_CONTEXT;
+    // this indicates that the device library has the test picture:
+    var isTestPictureSaved = false;
+    // this indicates that there was a critical error and tests cannot continue:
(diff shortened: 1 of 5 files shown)


appium-tests/android/android.spec.js
```diff
@@ -1,28 +1,34 @@
...
 var wd = wdHelper.getWD();
 var cameraConstants = require('../../www/CameraConstants');
 var cameraHelper = require('../helpers/cameraHelper');
+var screenshotHelper = require('../helpers/screenshotHelper');
 
 describe('Camera tests Android.', function () {
+
+    var STARTING_MESSAGE = 'Ready for action!',
+        RETRY_COUNT = 3, // how many times to retry taking a picture before failing
+        MINUTE = 60 * 1000;
```
dblotsky: Please put these outside of function scope. Nitpick: please have one declaration per line.

appium-tests/android/android.spec.js
```diff
@@ -1,28 +1,34 @@
...
-        // this indicates if device library has test picture:
+        // this indicates if device library has the test picture:
         isTestPictureSaved = false,
-        // this indecates that there was critical error and tests cannot continue:
+        // this indecates that there was a critical error and tests cannot continue:
         stopFlag = false,
-        // we need to know the screen width and height to properly click on the first image in the gallery
+        // we need to know the screen width and height to properly click on an image in the gallery
         screenWidth = 360,
         screenHeight = 567;
```
dblotsky: Should these guys also be constants?

appium-tests/ios/ios.spec.js
```diff
@@ -8,8 +14,12 @@ var wd = wdHelper.getWD();
 var isDevice = global.DEVICE;
 var cameraConstants = require('../../www/CameraConstants');
 var cameraHelper = require('../helpers/cameraHelper');
+var screenshotHelper = require('../helpers/screenshotHelper');
 
 describe('Camera tests iOS.', function () {
+
+    var MINUTE = 60 * 1000;
```
dblotsky: Please put this at the top of the file, outside of function scope.

appium-tests/android/android.spec.js
```diff
@@ -229,7 +255,8 @@ describe('Camera tests Android.', function () {
                             pass: false,
                             message: msg
                         };
-                        if (msg.indexOf('Error response status: 6') >= 0) {
+                        if (msg.indexOf('Error response status: 6') >= 0 ||
+                            msg.indexOf('Error response status: 7') >= 0) {
```
dblotsky: Please comment the significance of these strings.

appium-tests/android/android.spec.js
```diff
@@ -1,28 +1,34 @@
-/*jslint node: true, plusplus: true */
+/*jshint node: true */
 /*global beforeEach, afterEach */
```
vladimir-kotikov: Minor nit: this might be easier to use something like `/*jshint node:true, jasmine: true*/` to avoid declaring all Jasmine globals

appium-tests/android/android.spec.js
```diff
@@ -295,13 +318,35 @@ describe('Camera tests Android.', function () {
...
                     .finally(done);
             }
             done();
-        }, 600000);
+        }, 3 * MINUTE);
+
+        afterEach(function (done) {
+            // recreate the session if there was a critical error in the spec
+            if (stopFlag) {
+                return driver
```
vladimir-kotikov: This is not obvious that you're `return`'ing a promise here just to avoid second `done` call at the end of `afterEach`

I'd recommend you to reverse `if` condition to enclose the shortest code path and reduce nesting, and to explicitly do a return after `done`. Something like:
```javascript
if (!stopFlag) {
    done();
    return;
}
// promise chain here 
```

appium-tests/android/android.spec.js
alsorokin: No, these values can change when we determining the real device resolution.
These are just default values.

appium-tests/android/android.spec.js
```diff
@@ -1,28 +1,34 @@
...
+
     var driver,
-        startingMessage = 'Ready for action!',
         // the name of webview context, it will be changed to match needed context if there are named ones:
         webviewContext = 'WEBVIEW',
-        // this indicates if device library has test picture:
+        // this indicates if device library has the test picture:
         isTestPictureSaved = false,
-        // this indecates that there was critical error and tests cannot continue:
+        // this indecates that there was a critical error and tests cannot continue:
```
dblotsky: "indecates" -> "indicates"

dblotsky:
Thanks for fixing some of the grammar errors in the comments, btw!

appium-tests/android/android.spec.js
dblotsky: Please then set the defaults to be constants at the top just so we don't have seemingly magic values in the code.

PR #178, opened 2016-02-23
fix(android): Create Pictures directory if it doesn't exist

TheBosZ:
Possibly should throw an exception if the directory can't be created. A log message is emitted and behavior is no worse than current code.

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -550,6 +550,9 @@ private String getPicutresPath()
     String imageFileName = "IMG_" + timeStamp + (this.encodingType == JPEG ? ".jpg" : ".png");
     File storageDir = Environment.getExternalStoragePublicDirectory(
             Environment.DIRECTORY_PICTURES);
+    if (!storageDir.exists() && !storageDir.mkdir()) {
+        Log.e(LOG_TAG, "Error creating directory: " + storageDir.toURI().toASCIIString());
+    }
     String galleryPath = storageDir.getAbsolutePath() + "/" + imageFileName;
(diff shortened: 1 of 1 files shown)


infil00p:
Which device doesn't have this folder? Also, for some reason this failed Travis.  Can you clean up the conficts, create a JIRA issue and provide more details for the repro? Thanks!

TheBosZ:
It probably failed because this branch is out of date.

I can't for the life of me remember the exact device this happened to.

It's a very straightforward fix, but I've moved on. If you don't want/need this, feel free to close.

filmaj:
FWIW I had this problem when using an Android emulator but forgot to specify the SD card size in its AVD definition. In that situation, the emulator acts like the SD card is mounted in read-only mode and thus the `/storage/sdcard` mount is empty (and lacks the expected `./Pictures` subdirectory).

DavidKuennen:
We constantly have this error whenever we update this plugin (we have to modify the plugin ourself to create the folder if it doesn't exist). It's be awesome if this would be fixed in the original.

TheBosZ:
@DavidKuennen It's really pretty easy to fork this plugin, update the code and then just specify the fork in your config.xml.

That's what I do when I need a change to a plugin that the authors don't want to incorporate.

DavidKuennen:
Oh, I didn't know that! Thanks die Tage Info.

Am 17.05.2017 5:40 nachm. schrieb "Nathan" <notifications@github.com>:

jcesarmobile:
current code is calling mkdirs(), so no need for this

PR #180, opened 2016-02-26
Adding focus handler to make sure filepicker gets launched when app i…

rakatyal:
…s active

FilesavePicker gives error when launched from an app which is currently not active. When the capture is complete after launching CameraUI (the app becomes inactive when cameraUI is launched), we immediately launch filesavepicker introducing a race condition, in which sometimes the savepicker gets launched when the app hasn't reached the active state yet causing it to break.
We should only launch it once the app becomes active again.

### DIFF
--- a/src/windows/CameraProxy.js
+++ b/src/windows/CameraProxy.js
@@ -731,21 +731,32 @@ function takePictureFromCameraWindows(successCallback, errorCallback, args) {
 
     cameraCaptureUI.photoSettings.maxResolution = maxRes;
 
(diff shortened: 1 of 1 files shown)


dblotsky:
LGTM.

daserge:
@rakatyal, looks like JSHint has discovered some [issues](https://travis-ci.org/apache/cordova-plugin-camera/builds/112161028) with the code.

rakatyal:
@daserge: Thanks for pointing out. Making the needed changes.

rakatyal:
Yup it does break on Windows 8.1. Will fix it soon. Thanks for bringing this to our attention.

PR #181, opened 2016-02-26
CB-10711: getPicture() should NOT save to photo album when called wit…

omefire:
…h destinationType set Camera.DestinationType.NATIVE_URI


omefire:
I just realized that unless the image is saved to the photo album, we won't be able to have a native URI for iOS(ms-assets://). 
So, I'll change this PR to have a comment to document this limitation.

PR #183, opened 2016-03-01
Fix for CB-10050 : Plugin Camera : Inconsistent saveToPhotoAlbum beha…

ThillaiganeshChan:
…vior on PHOTOLIBRARY images

If sourceType == UIImagePickerControllerSourceTypePhotoLibrary
Then saveToPhotoAlbum is set False Always.


riknoll:
@omefire can you take a look at this?

omefire:
I think this change is not necessary because the user could have edited the picture they picked from the PHOTOLIBRARY and want it saved (even though it ends up being saved twice). This verification, in my opinion, is better left to the user of the plugin.

PR #184, opened 2016-03-01
CB-10750 Appium tests: fail fast if session is irrecoverable

alsorokin:
https://issues.apache.org/jira/browse/CB-10750


alsorokin:
Merging this now because it will help to lessen the load on slaves (currently 14 builds are pending)

PR #185, opened 2016-03-01
[iOS] Major refactor

dieppe:
This is a rebased version of the previous pull-request https://github.com/apache/cordova-plugin-camera/pull/101


src/ios/CDVCamera.h
```diff
@@ -87,19 +91,51 @@ typedef NSUInteger CDVMediaType;
...
 
 /*
- * getPicture
+ * takePicture
  *
  * arguments:
- *	1: this is the javascript function that will be called with the results, the first parameter passed to the
- *		javascript function is the picture as a Base64 encoded string
- *  2: this is the javascript function to be called if there was an error
+ *  1: success callback (called with either Base64, file URI or native URI)
```
omefire: These comments seem to be meant to document the javascript side of things, instead of the objective-c side. What's the reason for them being here ?

src/ios/CDVCamera.m
```diff
@@ -59,6 +59,59 @@ Licensed to the Apache Software Foundation (ASF) under one
 
 @implementation CDVPictureOptions
 
+/*
+ API:
+ Camera:
+ - CameraPopoverHandle:
```
omefire: it's a bit hard to follow these options since there's no indentation.
for example, is 'getPicture' also part of 'CameraPopoverHandle' ?

omefire: Also, why do we have the API documented here ? If the documentation is lacking, shouldn't we add it to the README.md that shows up so that app developers can also take advantage of it ?

src/ios/CDVCamera.m
```diff
@@ -59,6 +59,59 @@ Licensed to the Apache Software Foundation (ASF) under one
...
+ 1: ARROW_UP
+ 2: ARROW_DOWN
+ 4: ARROW_LEFT
+ 8: ARROW_RIGHT
+ 15: ARROW_ANY
+ - cleanup => ()
+ */
+
+/*
+ TODO:
```
omefire: Please, File JIRA issues for these ToDO's

src/ios/CDVCamera.m
```diff
@@ -78,7 +131,10 @@ + (instancetype) createFromTakePictureArguments:(CDVInvokedUrlCommand*)command
     pictureOptions.mediaType = [[command argumentAtIndex:6 withDefault:@(MediaTypePicture)] unsignedIntegerValue];
     pictureOptions.allowsEditing = [[command argumentAtIndex:7 withDefault:@(NO)] boolValue];
     pictureOptions.correctOrientation = [[command argumentAtIndex:8 withDefault:@(NO)] boolValue];
-    pictureOptions.saveToPhotoAlbum = [[command argumentAtIndex:9 withDefault:@(NO)] boolValue];
+
+    BOOL isDestinationNativeUri = (pictureOptions.destinationType == DestinationTypeNativeUri);
+    BOOL isSourceCamera = (pictureOptions.sourceType == UIImagePickerControllerSourceTypeCamera);
+    pictureOptions.saveToPhotoAlbum = [[command argumentAtIndex:9 withDefault:@(NO)] boolValue] || (isDestinationNativeUri && isSourceCamera);
```
omefire: A comment for the reason behind this 'quirk' would be good.

src/ios/CDVCamera.m
```diff
@@ -145,9 +201,38 @@ - (void)takePicture:(CDVInvokedUrlCommand*)command
     [self.commandDelegate runInBackground:^{
         
         CDVPictureOptions* pictureOptions = [CDVPictureOptions createFromTakePictureArguments:command];
+        
+        /*
+         FIXME #1
+         What to do about quality?
```
omefire: Let's create a JIRA issue for this one as well.

omefire: I think if the quality option is specified, we need to modify the image retrieved from the PhotoLibraryr or SavedAlbum & (re)save it with the specified quality, then send a 'reference' back to the newly saved image.
This looks like a potential change in the API and might need to be ran past the dev mailing list.

src/ios/CDVCamera.m
```diff
@@ -145,9 +201,38 @@ - (void)takePicture:(CDVInvokedUrlCommand*)command
...
+         If the option is set and the image is retrieved from the PhotoLibrary or SavedAlbum with a NATIVE_URI, no editing,
+         and is not saved to the photo album, then we return the "wrong" image.
+         However, the doc says "quality of the **saved** image", so it might be fine.
+         */
+        /*
+         FIXME #2
+         The doc says: "Rotate the image to correct for the orientation of the device **during capture**."
```
omefire: Yes. and possibly, it's also used for images picked from photolibrary or savedalbum, not sure though.

src/ios/CDVCamera.m
```diff
@@ -145,9 +201,38 @@ - (void)takePicture:(CDVInvokedUrlCommand*)command
...
+        BOOL isDestinationNativeUri = (pictureOptions.destinationType == DestinationTypeNativeUri);
+        
+        BOOL needsResize = [self needsResize:pictureOptions];
+        BOOL needsOrientationCorrection = pictureOptions.correctOrientation;
+        BOOL isSourceCamera = (pictureOptions.sourceType == UIImagePickerControllerSourceTypeCamera);
+        BOOL allowsEditing = pictureOptions.allowsEditing;
+        BOOL needsSavingToPhotoAlbum = (needsResize || needsOrientationCorrection || allowsEditing);
+        
+        // if one wants an edited image and a NATIVE_URI, the edited image must be in the assets library therefore one must set the saveToPhotoAlbum option to true.
+        if (!pictureOptions.saveToPhotoAlbum && isDestinationNativeUri && needsSavingToPhotoAlbum) {
```
omefire: We also need to account for when the 'quality' option is not set the DEFAULT value. That also counts as an edit.

omefire: Hmmm, I agree that we should ideally not save into the photoAlbum if the user hasn't specified it in the options.
But on the other hand, the current behavior is to edit the image anyway, then go ahead and save it in the assets library/saved photo album anyway.
Which one should we go with ? I'm not really sure.

src/ios/CDVCamera.m
```diff
@@ -145,9 +201,38 @@ - (void)takePicture:(CDVInvokedUrlCommand*)command
...
+            NSLog(@"Incompatible options, cannot return native URI if image is not in the assets library");
+            CDVPluginResult* result = [CDVPluginResult resultWithStatus:CDVCommandStatus_ERROR messageAsString:@"Incompatible options, cannot return native URI if image is not in the assets library"];
+            [weakSelf.commandDelegate sendPluginResult:result callbackId:command.callbackId];
+            return;
+        }
+        
         pictureOptions.popoverSupported = [weakSelf popoverSupported];
         pictureOptions.usesGeolocation = [weakSelf usesGeolocation];
-        pictureOptions.cropToSize = NO;
```
omefire: +1

src/ios/CDVCamera.m
```diff
@@ -402,6 +444,115 @@ - (NSString*)tempFilePath:(NSString*)extension
     return filePath;
 }
 
+// This is not absolute, but more mightNeedOrientationCorrection
+- (BOOL) needsOrientationCorrection:(UIImage*)image options:(CDVPictureOptions*)options
```
omefire: We don't need a separate function for this, I think we can have these options 'inline' where needed instead of creating separate functions for them.

src/ios/CDVCamera.m
```diff
@@ -402,6 +444,115 @@ - (NSString*)tempFilePath:(NSString*)extension
...
+    return (options.targetSize.height > 0 && options.targetSize.width > 0);
+}
+
+- (BOOL) needsEdit:(UIImage*)image options:(CDVPictureOptions*)options
+{
+    return [self needsOrientationCorrection:image options:options] || [self needsResize:options];
+}
+
+- (BOOL) needsSavingToPhotoAlbum:(UIImage*)image options:(CDVPictureOptions*)options
+{
```
omefire: If the user specifies the option saveToPhotoAlbum and the image already exists in the PhotoLibrary, shouldn't we save it to the SavedPhotoAlbum ? (since SavedPhotoAlbum is different from the PhotoLibrary).

src/ios/CDVCamera.m
```diff
@@ -402,6 +444,115 @@ - (NSString*)tempFilePath:(NSString*)extension
...
+    BOOL saveToPhotoAlbum = options.saveToPhotoAlbum && ([self needsEdit:image options:options] || isSourceCamera);
+    
+    return saveToPhotoAlbum;
+}
+
+/*
+ Metadata is not needed for:
+ - source: gallery
+ - destination: NATIVE_URI
+ - no edit (not orientation and not resize and not allowEdits)
```
omefire: what about 'quality' changes ?

src/ios/CDVCamera.m
```diff
@@ -402,6 +444,115 @@ - (NSString*)tempFilePath:(NSString*)extension
...
+ 
+ otherwise, it can be found:
+ - source: camera
+ => UIImagePickerControllerMediaMetadata
+ - source: gallery
+ => CGImage thingy
```
omefire: what 'thingy' ?

src/ios/CDVCamera.m
```diff
@@ -344,49 +429,6 @@ - (void)popoverControllerDidDismissPopover:(id)popoverController
     self.hasPendingOperation = NO;
 }
 
-- (NSData*)processImage:(UIImage*)image info:(NSDictionary*)info options:(CDVPictureOptions*)options
```
omefire: [Just so I can remember to check what replaced the functionality that used to be provided by this funtion]

src/ios/CDVCamera.m
```diff
@@ -402,6 +444,115 @@ - (NSString*)tempFilePath:(NSString*)extension
...
+//      if (geoLoc)
+//          save on self.{data, metadata}
+//          wait for geoLoc
+//      else
+//          saveAndResult
+//
+// saveAndResult:
+//      saveToPhotoAlbum if needed (options.saveToPhotoAlbum && !(gallery && NATIVE_URI))
+//      send back the image URI
+- (void)didReceiveImage:(CDVPictureOptions*)options info:(NSDictionary*)info
```
omefire: For consistency's sake, use the same naming convention for both image results and video results. (cf: 'resultForVideo')

src/ios/CDVCamera.m
dieppe: Afaict, quality is not stored in metadata

src/ios/CDVCamera.m
dieppe: Sorry, forgot to remove those comments. They were only meant to remember how to access metadata in different situation (and might be totally outdated).

src/ios/CDVCamera.h
dieppe: I'll remove that

src/ios/CDVCamera.m
dieppe: This was meant as a quick way to check the API (and most of all the options) while developing. I can get rid of it (but it can be a good way to know if we're in sync with the docs).

src/ios/CDVCamera.m
dieppe: Sorry, this is a leftover comment, this TODOs are already implemented.

src/ios/CDVCamera.m
dieppe: Agreed. The only reason this is a function is because I wanted to check:
- if the image really needed an orientation correction (but that is an over optimization),
- if we really wanted to correct gallery image orientation.
Anyway, will be inlined.

src/ios/CDVCamera.m
dieppe: It's trivial to change. Waiting for your confirmation to do so. In general, I am against duplicating images on the user's phone. But if you think it makes sense to allow it, fine by me :wink:.

dieppe:
Made some changes to comply with your comments

omefire:
Thanks @dieppe , I'll resume looking at this PR soon. I've been out.

src/ios/CDVCamera.m
```diff
@@ -145,9 +198,38 @@ - (void)takePicture:(CDVInvokedUrlCommand*)command
...
+        
+        BOOL needsResize = [self needsResize:pictureOptions];
+        BOOL needsOrientationCorrection = pictureOptions.correctOrientation;
+        BOOL isSourceCamera = (pictureOptions.sourceType == UIImagePickerControllerSourceTypeCamera);
+        BOOL allowsEditing = pictureOptions.allowsEditing;
+        BOOL needsSavingToPhotoAlbum = (needsResize || needsOrientationCorrection || allowsEditing);
+        
+        // if one wants an edited image and a NATIVE_URI, the edited image must be in the assets library therefore one must set the saveToPhotoAlbum option to true.
+        if (!pictureOptions.saveToPhotoAlbum && isDestinationNativeUri && needsSavingToPhotoAlbum) {
+            NSLog(@"Incompatible options, cannot return native URI if image is not in the assets library");
```
omefire: +1

src/ios/CDVCamera.m
```diff
@@ -145,9 +198,38 @@ - (void)takePicture:(CDVInvokedUrlCommand*)command
     [self.commandDelegate runInBackground:^{
```
omefire: Never mind, I see it below.

src/ios/CDVCamera.m
```diff
@@ -398,10 +437,82 @@ - (NSString*)tempFilePath:(NSString*)extension
...
+    /*
+     We can send the result immediately if:
+     - we fetch the picture from the PhotoLibrary or the SavedPhotoAlbum,
+     - we don't do any editing (orientation or resize),
+     - we pass down the result as a NATIVE_URI
+     */
+    BOOL needsEdit = [self needsEdit:image options:options];
+    BOOL isSourceCamera = (options.sourceType == UIImagePickerControllerSourceTypeCamera);
+    BOOL isDestinationNativeUri = (options.destinationType == DestinationTypeNativeUri);
+    BOOL needsMetadata = (needsEdit || isSourceCamera || !isDestinationNativeUri);
```
omefire: I feel like the description given in the 'Metadata' section here could help clarify things better here, can we add it ?

https://github.com/apache/cordova-plugin-camera/pull/101

"Regarding metadata, we do not need it in only one case, when the destination type is NATIVE_URI, the source is the library, and no edits have to be done."

infil00p:
Closing this PR, since it eventually got abandoned.  I'm not quite up to speed on iOS dev to know what this does, but I do know that it's conflicting. :(

PR #188, opened 2016-03-02
CB-10414: Adding focus handler to resume video when user comes back o…

rakatyal:
…n leaving the app while preview was running


sarangan12:
LGTM

PR #189, opened 2016-03-04
Fix Crash on Android 6.0 for permissions

vishalkardode:
Removed check `fileWillBeModified` before requesting runtime Permissions.
When you don't specify image height / width for picker, plugin does not ask for `READ_EXTERNAL_STORAGE `permission and results in crash while file selection.


riknoll:
This was just fixed in a19c752

PR #191, opened 2016-03-09
CB-10825 android: Always request READ permission for gallery source

riknoll:
I am unhappy with this change, but this issue is blocking a release and a proper fix could end up being a major rewrite to how we do permissions. The issue at the center of this is that whether or not we need the READ permission changes according to the camera options passed to `getPicture()` and the app used to select a photo. We need to fix how we request permissions so that we request them *after* the photo is selected, and only when we need them. 

Really, the bigger issue is that we try and convert all `content://` URIs into real file paths. As I understand it, Content resolvers are not intended for that purpose.

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -168,11 +168,8 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
                     this.callTakePicture(destType, encodingType);
                 }
                 else if ((this.srcType == PHOTOLIBRARY) || (this.srcType == SAVEDPHOTOALBUM)) {
-                    // Any options that edit the file require READ permissions in order to try and
-                    // preserve the original exif data and filename in the modified file that is
-                    // created
-                    if(this.mediaType == PICTURE && (this.destType == FILE_URI || this.destType == NATIVE_URI)
-                            && fileWillBeModified() && !PermissionHelper.hasPermission(this, permissions[0])) {
+                    // FIXME: Stop always requesting the permission
+                    if(!PermissionHelper.hasPermission(this, permissions[0])) {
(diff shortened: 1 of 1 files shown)


riknoll:
@infil00p @csantanapr @jcesarmobile please review

csantanapr:
@riknoll I'm not too familiar with android permission would be up to @infil00p @jcesarmobile or maybe @macdonst

macdonst:
@riknoll if you call getPicture() to read from the gallery shouldn't it request READ permission before opening the gallery?

riknoll:
I guess it does make more sense for us to request the permission before we select the photo because the developer of a Cordova app doesn't really have a way to gracefully respond to permission denials after the photo has been selected.

That being said, we don't actually need the permission to get the photo because we use an Intent to launch the native gallery app. The reason we request the permission is because sometimes the stuff we do with the URI we get back from the native Activity requires us to have permissions. I feel that the fact that we are using intents means that we should not be requesting permissions at all. If we have to keep requesting permission to match the behavior of other platforms, it would be nice if we could be a bit more granular about it and avoid it when possible. That approach is currently thwarted by `getRealPath()`, which sometimes requires permissions based on the type of URI it is passed (which is dependent on the gallery app the user chooses and thus cannot be predicted beforehand)

macdonst:
@riknoll yeah, we shouldn't let people into the gallery if we can't read the file later.

If we can't tell which gallery they will use is there anyway around this?

riknoll:
@macdonst Not with the current implementation of `getRealPath()`

riknoll:
I think that fixing (or removing) `getRealPath()` might be a large undertaking, so it shouldn't block the release. This permission request at least makes sense to the user of the app.

csantanapr:
@riknoll what's the final conclusion for the plugin release?

riknoll:
Alright, I'm going to merge this in

PR #194, opened 2016-03-11
CB-10460 getRealPath return null in some cases

jcesarmobile:
Now there is only a function to get the real path on API 11 and above.


sassy:
I check the this patch the following device. (try to run https://github.com/sassy/cordovaCameraPluginFailSample )

* Motorola Xoom (Android 4.0.3) OK
* Xperia Z2 Tablet (Android 4.4.2) OK
* Nexus9 (Android 5.0) OK

sassy:
I checked it used cordova-plugin-camera-tests.
Xperia Z2 Tablet (Android 4.4.2) is no problem.

PR #196, opened 2016-03-23
Making focus handlers to only work for Windows 10 phone

rakatyal:
The focus handler behave differently on Win 8.1, so making the change to  work only in case for windows 10 phone.

### DIFF
--- a/src/windows/CameraProxy.js
+++ b/src/windows/CameraProxy.js
@@ -745,8 +745,9 @@ function takePictureFromCameraWindows(successCallback, errorCallback, args) {
     cameraCaptureUI.photoSettings.maxResolution = maxRes;
 
     var cameraPicture;
-    var savePhotoOnFocus = function() {
-
+    
+    // define focus handler for windows phone 10.0
+    var savePhotoOnFocus = function () {
         window.removeEventListener("focus", savePhotoOnFocus);
         // call only when the app is in focus again
         savePhoto(cameraPicture, {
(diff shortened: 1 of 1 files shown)


src/windows/CameraProxy.js
```diff
@@ -758,16 +759,31 @@ function takePictureFromCameraWindows(successCallback, errorCallback, args) {
         }, successCallback, errorCallback);
     };
 
-    // add and delete focus eventHandler to capture the focus back from cameraUI to app 
-    window.addEventListener("focus", savePhotoOnFocus);
-    cameraCaptureUI.captureFileAsync(WMCapture.CameraCaptureUIMode.photo).done(function(picture) {
+    // if windows phone 10, add and delete focus eventHandler to capture the focus back from cameraUI to app
+    if (navigator.appVersion.indexOf('Windows Phone 10.0') >= 0) {
```
jasongin: Is this the best/only way to check this? It will fail on future versions of Windows Phone if the version number changes to something like 10.1 or 11.0.

And what about non-phone Windows 10?

rakatyal: I think this is our best option. Also the other checks in this file seems to use this property as well. It would break when the version changes though yes. 
Non phone windows 10 works fine without the focus handler.

jasongin: OK.

LGTM

rakatyal:
This has been merged.

PR #198, opened 2016-03-29
CB-10857 android : Camera.getPicture return null for Google Drive (camera 2.1.1)

elsewhat:
FileHelper getRealPath will fallback to URI string if no other type of
path has been found

### DIFF
--- a/src/android/FileHelper.java
+++ b/src/android/FileHelper.java
@@ -59,6 +59,10 @@ public class FileHelper {
         else
             realPath = FileHelper.getRealPathFromURI_API11_And_Above(cordova.getActivity(), uri);
 
+        //Return the URI string if no other type of path has been found
+        if(realPath == null)
+            realPath =  uri.toString();
+
         return realPath;
     }
 


elsewhat:
Pull request has been tested by debugging. The cordova-plugin-camera plugin was added to a local cordova project from my forked repository and the new commit.

riknoll:
So, I've done a bit of testing on Marshmallow and it seems like this works pretty well. It does have the caveat of returning a `content://` URI, but that isn't a huge deal in my book. Currently, depending on the camera options you pass it is possible to get back either a `content://` URI, `file://` URI, or a file path back, so we aren't consistent at all anyway.

@jcesarmobile does this look alright to you? I plan to do some more testing on Jelly Bean in a bit.

jcesarmobile:
Sorry, I've been really busy this week and didn't have time to test
The code changes LGTM, but I want to test it tomorrow, and I think we should document that google drive will return a content:// url, and maybe the limitations of a content:// url

jcesarmobile:
Tested that the images are correctly displayed if used on an img tag, and can be uploaded to a server using file-transfer.

@riknoll, you are right, the paths we return aren't very consistent, maybe we should review them or at least document which params return which kind or path.

But for now, this looks good and I think it can be merged.

riknoll:
I think that we should eventually fix it to be consistent (and figure out the meaning of NATIVE_URI and FILE_URI for that matter). Also, we should definitely make sure it always returns a URI and not a path

riknoll:
Okay, finally got around to testing this on Jelly bean and I found an issue: content URIs don't always end in a file type (like .jpg). We try and get the filename in `ouputModifiedBitmap` from `getRealPath` (see [here](https://github.com/elsewhat/cordova-plugin-camera/blob/CB-10857cordova-plugin-camera/src/android/CameraLauncher.java#L600)) and so we end up writing to a temporary image without a file extension. I don't know if we make assumptions like that anywhere else, but we need to address that before merging this. Reproducing camera options:
```
{
        mediaType: Camera.MediaType.PICTURE,
        destinationType: Camera.DestinationType.FILE_URI,
        sourceType: Camera.PictureSourceType.SAVEDPHOTOALBUM,
        quality: 50,
        allowEdit: false,
        encodingType: 0,
        correctOrientation: true
        saveToPhotoAlbum: true
}
```

jcesarmobile:
Ok, so we should change 
```
```
As realPath will not be null anymore it doesn't make any sense.

We have some options:
Detect if the fileName have an extension, if not, then use the "modified"... approach.
Try to get the real image name, I think that's possible even for Google Drive images.

Any other ideas?

riknoll:
If we can get the real image name, we should do that. However, since we are modifying it, we should also mark it as such and put `-modified` at the end or something. I've also noticed some weird behavior when we always use the same filename after modifying images, so we might need to timestamp those files as well. We just need to make sure the cache is cleared on a regular basis, I forget if Android handles that automatically.

jcesarmobile:
I like the idea, if we can get the real name, then the new image should be realname-modified-timestap, if we can't get the real name, then just use modified-timestamp.

About the cache, it seems that we should clean it.

> When the device is low on internal storage space, Android may delete these cache files to recover space. However, you should not rely on the system to clean up these files for you. You should always maintain the cache files yourself and stay within a reasonable limit of space consumed, such as 1MB. When the user uninstalls your application, these files are removed.


http://developer.android.com/intl/es/guide/topics/data/data-storage.html#ExternalCache

riknoll:
That sounds good to me. We can probably just put the images into a directory like `cache/cordova-plugin-camera` and routinely wipe it as part of `onCreate` or something. Just need to make sure people know to use `saveToPhotoAlbum=true` if they need an image persistently stored.

elsewhat:
@riknoll @jcesarmobile thanks for the input. What's the way forward for this pull request?

The your findings requires a bit to much knowledge of the plugin for me to handle at the current time. Is it an idea to merge this pull request and open the other findings in a new issue? After all this pull request fixes an important issue in the current version of the plugin

ajaysaini95700:
it is working fine but i am not able to upload image from google photos . whenever i try to upload image it returns null path of image .

harellevy:
is this working with video from google drive?
when is this going to be merged to the master?

upload videos from local directory in my phone works fine but when i try upload videos from google drive it returns null as the success result

claudioviola:
Is there any news about this pull request?
My version is 2.3.0 and I see the bug. At moment for fix it the only way is to download this and merge manually? 
https://github.com/apache/cordova-plugin-camera/pull/198/commits/b8b9b5abced904f1b0990eef6ee8acf3e6b44987

mglendrange:
I am also eagerly waiting for this pull request. Any progress?

infil00p:
Compatibility work takes priority over this fix.  Once that fix is merged, I'll look at adding this, although I believe another patch has already addressed this.

ppiacez:
Any news?

powermosfet:
We're still seeing this issue in 2.4.1. Uploads from Google Drive return the path ""

esase:
I have the same issue

jd0048:
Same issue with dropbox also.

mglendrange:
Since there was no progress here we have moved away from using this plugin to handle uploads. Instead we trigger a click on a hidden HTML input field with type set to file. You need to add an event listener that listens to the change event and use a FileReader to handle the change event.

Works almost perfect for both iOS and Android. Only issue is accessing OneDrive, where nothing is returned. 

For taking photos we still use this plugin on Android as the camera would not trigger when clicking the input field, but for iOS we are also doing this now by triggering a click on the input field.

jcesarmobile:
https://github.com/apache/cordova-plugin-camera/issues/391

jcesarmobile:
This is working fine now on master branch (4.1.0-dev), so going to close the PR and issue

kareemyousry:
it return null when i select video from google drive ,but it work fine with image

MatthDefrenne:
It return null as well on lastest version

PR #199, opened 2016-03-29
Update plugin.xml

mehmetcodes:
In principle Cordova should not use a permission where it is no longer relevant as it might leave things less secure.

You can see this documentation here http://developer.android.com/guide/topics/manifest/uses-permission-element.html

Note the following in the documentation:
# 

android:maxSdkVersion
The highest API level at which this permission should be granted to your app. Setting this attribute is useful if the permission your app requires is no longer needed beginning at a certain API level.
For example, beginning with Android 4.4 (API level 19), it's no longer necessary for your app to request the WRITE_EXTERNAL_STORAGE permission when your app wants to write to its own application-specific directories on external storage (the directories provided by getExternalFilesDir()). However, the permission is required for API level 18 and lower. So you can declare that this permission is needed only up to API level 18 with a declaration such as this:

<uses-permission
     android:name="android.permission.WRITE_EXTERNAL_STORAGE"
     android:maxSdkVersion="18" />
This way, beginning with API level 19, the system will no longer grant your app the WRITE_EXTERNAL_STORAGE permission.
# This attribute was added in API level 19.

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -65,7 +65,7 @@
             </feature>
         </config-file>
         <config-file target="AndroidManifest.xml" parent="/*">
-            <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+            <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" android:maxSdkVersion="18"/>
         </config-file>
 
         <source-file src="src/android/CameraLauncher.java" target-dir="src/org/apache/cordova/camera" />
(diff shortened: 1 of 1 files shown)


riknoll:
I believe we do actually require that permission even in API level 19+. The current code requires access to the photos directory, which is outside of the application storage.

mehmetcodes:
You mean to say if we compile, contrary to the docs API 19+ still demands this permission?

Well, anyways what I can say is that I noticed that when multiple plugins have the permission, because the attributes are different, they all end up in the Android manifest.  Before applying this code, you'll need a strategy for handling attributes in permissions baked into the plugin.xml parser.  That'll take time.

PR #201, opened 2016-04-07
CB-11034: Camera.getPicture -Handle of OutOfMemoryError gracefull

elsewhat:
Catch OutOfMemoryError for memory intensive operations and handle them
gracefully.

Tested on a Samsung SII which throws OutOfMemoryException when taking a photo and fetching it as a DATA_URL.Not able to test when selecting from gallery

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -26,6 +26,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.lang.OutOfMemoryError;
 import java.net.URI;
 import java.text.SimpleDateFormat;
 import java.util.Date;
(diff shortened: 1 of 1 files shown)


nikhilkh:
Silently handling OutOfMemoryError is not a good idea - we should look to optimize the code so that it uses less memory instead.

riknoll:
Ah, I think I misunderstood your JIRA and I agree with @nikhilkh that optimizing our memory use is the right approach. Looking through the code, it seems we already consume that exception in a number of places when dealing with bitmaps which is probably a very bad thing. Google provides some guidance [here](http://developer.android.com/training/displaying-bitmaps/manage-memory.html) on managing bitmap memory.

riknoll:
There are definitely cases where native exceptions should be surfaced to the javascript, but memory issues are the responsibility of the native developer. There isn't much that can be done from the Webview to correctly handle that situation

elsewhat:
@riknoll @nikhilkh as an app developer I have to disagree with you.
I would **expect a framework such as cordova to never crash an app under no circumstances if it's preventable**.
Crashing an app is a very significant event which impacts users perception of the app very negatively and causes a significant effort wrt. end-user support for enterprise apps. It will in many cases cause entered data to be lost, and nothing frustrates users more.

There is lots the app can do to handle such an exception. In my case I can make sure the users understand the issue is because the size of the image and how they can perform workarounds. 

Ideally, the CameraLauncher.failPicture would allow me to pass some semantics regarding the nature of the issue for example an error code. But since this was not in place already in the code, I didn't do any changes to it. 

However, **I totally agree that refactoring the code to reduce the memory footprint is an even better approach which actually tackles the root cause. But this doesn't negate the need to avoid OutOfMemoryErrors when performing memory intensive operations**. Even with an exceptionally optimized memory handling of images in cordova-camera-plugin there will always be situations where an OutOfMemoryError occurs, for example if the app uses a lot of memory on the webview side and the device is old with a strict limit on heap size (a G1 had 16MB of heap)

PR #202, opened 2016-04-07
Appium tests stability improvements

alsorokin:


riknoll:
I'm not sure if this is a limitation of the way appium injects functions, but since the promise-related callbacks seem to be the same for iOS and Android, can they be moved to a common location?

appium-tests/android/android.spec.js
```diff
@@ -450,143 +306,110 @@ describe('Camera tests Android.', function () {
...
+                .fail(saveScreenshotAndFail)
                 .deviceKeyEvent(4)
-                .sleep(2000)
-                .deviceKeyEvent(4)
-                .sleep(2000)
                 .elementById('action_bar_title')
                 .then(function () {
                     // success means we're still in native app
                     return driver
                         .deviceKeyEvent(4)
```
riknoll: It would be nice to have constants clarifying these `deviceKeyEvent` numbers. I'm guessing that this corresponds to pressing the back button?

alsorokin:
@riknoll addressed both of your comments

appium-tests/android/android.spec.js
```diff
@@ -27,52 +28,47 @@
...
     // we need to know the screen width and height to properly click on an image in the gallery:
     var screenWidth = DEFAULT_SCREEN_WIDTH;
     var screenHeight = DEFAULT_SCREEN_HEIGHT;
+    // promise count to use in promise ID
+    var promiseCount = 0;
 
-    function win() {
-        expect(true).toBe(true);
+    function getNextPromiseId() {
+        return PROMISE_PREFIX + promiseCount++;
```
vladimir-kotikov: Nit: it would be better to write `PROMISE_PREFIX + (++promiseCount)`, then in `getCurrentPromiseId` you could just return `PROMISE_PREFIX + promiseCount`

appium-tests/android/android.spec.js
```diff
@@ -251,141 +200,50 @@ describe('Camera tests Android.', function () {
...
-            .elementById('info')
-            .getAttribute('innerHTML')
-            .then(function (html) {
-                if (html !== STARTING_MESSAGE) {
-                    screenWidth = Number(html);
-                }
+        return driver
+            .context(webviewContext)
+            .execute(function () {
+                return window.innerWidth;
```
vladimir-kotikov: Is it possible to combine requesting both `window.innerWidth` and `window.innerHeight` into one `execute` call? I think you can just return an object `{width: ..., height: ...}`

appium-tests/ios/ios.spec.js
```diff
@@ -85,204 +84,187 @@ describe('Camera tests iOS.', function () {
...
                         });
                 }
                 if (options.hasOwnProperty('sourceType') && options.sourceType === cameraConstants.PictureSourceType.SAVEDPHOTOALBUM) {
                     return driver
-                        .elementByXPath('//UIACollectionCell')
+                        .waitForElementByXPath('//UIACollectionCell', MINUTE / 2)
                         .click()
                         .then(function () {
-                            if (options.hasOwnProperty('allowEdit') && options.allowEdit === true) {
+                            if (options.allowEdit) {
```
vladimir-kotikov: Please invert this statement to reduce nesting. The code inside looks overly indented.

appium-tests/android/android.spec.js
dblotsky: Or even break up the update and the read into two lines, so it's easier to read.

PR #203, opened 2016-04-08
Added Sample section to the Camera plugin README

Mikejo5000:


README.md
```diff
@@ -526,3 +526,207 @@ Tizen only supports a `destinationType` of
...
+
+## Take a picture and get a FileEntry Object
+
+If you want to do something like copy the image to another location, or upload it somewhere using the FileTransfer plugin, you need to get a FileEntry object for the returned picture. To do that, call `window.resolveLocalFileSystemURL` on the file URI returned by the Camera app. If you need to use a FileEntry object, set the `destinationType` to `Camera.DestinationType.FILE_URI` in your CameraOptions object (this is also the default value).
+
+```js
+function setOptions(srcType) {
+    var options = {
+        quality: 20,
+        destinationType: Camera.DestinationType.FILE_URI,
```
riknoll: It might be helpful to clarify that `window.resolveLocalFileSystemURL` requires that you add the file plugin

README.md
```diff
@@ -526,3 +526,207 @@ Tizen only supports a `destinationType` of
...
+        // In this app, dynamically set the picture source, Camera or photo gallery
+        sourceType: srcType,
+        encodingType: Camera.EncodingType.JPEG,
```
riknoll: Is there a reason you show the `setOptions` implementation here but not in all of the above examples? Looks like the same implementation

README.md
```diff
@@ -526,3 +526,207 @@ Tizen only supports a `destinationType` of
...
+        // Do something with the FileEntry object, like write to it, upload it, etc.
+        // writeFile(fileEntry, imgUri);
+        console.log("got file: " + fileEntry.fullPath);
+        displayFileData(fileEntry.nativeURL, "Native URL");
+
+    }, function () {
+        // If you can't easily get the FileEntry (for example, no write access
+        // to Pictures library) copy to a new FileEntry.
+        copyToFile(imgUri);
+    });
```
riknoll: Is `displayFileData()` defined anywhere?

README.md
```diff
@@ -526,3 +526,207 @@ Tizen only supports a `destinationType` of
...
+}
+```
+
+In the example shown in the preceding code, you call the app's `copyToFile` function if you don't get a valid FileEntry object. The image URI returned from the Camera app should result in a valid FileEntry, but platform behavior may be different for files returned from the file picker.
+
```
riknoll: Why wouldn't you be able to access the pictures library? I guess in Android the user could deny the permission request. In that case, it would not be able to read the file either, though. Is this something that you experienced in testing? Might be related to the file plugin bug that was blocking writes to external storage on Android (that bug should be fixed in the next plugin release)

README.md
```diff
@@ -526,3 +526,207 @@ Tizen only supports a `destinationType` of
...
+## Take a Picture and Return Thumbnails (Resize the Picture)
+
+To get smaller images, you can return a resized image by passing both `targetHeight` and `targetWidth` values with your CameraOptions object. In this example, you resize the returned images to 100px (the aspect ratio is maintained, so 100px is either the height or width, whichever is greater).
+
+```js
+function openCamera(selection) {
+
+    var srcType = Camera.PictureSourceType.CAMERA;
+    var options = setOptions(srcType);
```
riknoll: Instead of saying "resize the returned images to 100px", perhaps you could rephrase it to something like "resize the returned image to fit in a 100px by 100px box"

README.md
```diff
@@ -526,3 +526,207 @@ Tizen only supports a `destinationType` of
...
+```js
+function setOptions(srcType) {
+    var options = {
+        quality: 20,
+        destinationType: Camera.DestinationType.FILE_URI,
+        // In this app, dynamically set the picture source, Camera or photo gallery
+        sourceType: srcType,
+        encodingType: Camera.EncodingType.JPEG,
+        mediaType: Camera.MediaType.PICTURE,
+        allowEdit: true,
```
riknoll: Why do we recommend this quality? I'm not arguing against it, just curious.

README.md
```diff
@@ -526,3 +526,207 @@ Tizen only supports a `destinationType` of
...
+
+    if (selection == "picker-thmb") {
+        // To downscale a selected image,
+        // Camera.EncodingType (e.g., JPEG) must match the selected image type.
+        options.targetHeight = 100;
+        options.targetWidth = 100;
+    }
+
+    navigator.camera.getPicture(function cameraSuccess(imageUri) {
+
```
riknoll: What's the reason for the `Camera.EncodingType` comment? I think it should work either way (on Android at least). Is it different for other platforms?

riknoll:
I made a few comments inline, but nothing major.

Also, `README.md` is actually a generated file in this repo (and this repo only). It would be great if you could move the changes you made in that file over to `jsdoc2md/TEMPLATE.md`. `README.md` should get automatically regenerated with your changes after you make a commit. You can edit `README.md` directly in the other repos (the camera plugin is just special)

jsdoc2md/TEMPLATE.md
```diff
@@ -203,3 +203,194 @@ Tizen only supports a `destinationType` of
...
+
+    }, options);
+}
+```
+
+## Take a picture and get a FileEntry Object <a name="convert"></a>
+
+If you want to do something like copy the image to another location, or upload it somewhere using the FileTransfer plugin, you need to get a FileEntry object for the returned picture. To do that, call `window.resolveLocalFileSystemURL` on the file URI returned by the Camera app. If you need to use a FileEntry object, set the `destinationType` to `Camera.DestinationType.FILE_URI` in your CameraOptions object (this is also the default value).
+
+>*Note* You need the File plugin to call `window.resolveLocalFileSystemURL`.
```
riknoll: Can you link to the file plugin here?

jsdoc2md/TEMPLATE.md
```diff
@@ -203,3 +203,194 @@ Tizen only supports a `destinationType` of
...
+
+```js
+function copyToFile(imgUri) {
+    window.resolveLocalFileSystemURL(cordova.file.cacheDirectory, function success(dirEntry) {
+
+        // JPEG file
+        dirEntry.getFile("tempFile.jpeg", { create: true, exclusive: false }, function (fileEntry) {
+
+            // Do something with it, like write to it, upload it, etc.
+            // writeFile(fileEntry, imgUri);
```
riknoll: Can we have the full file writing implementation here to? Or can you link to the file plugin README where there is an example? Right now this function doesn't really show the copying part

jsdoc2md/TEMPLATE.md
```diff
@@ -203,3 +203,194 @@ Tizen only supports a `destinationType` of
...
+    }, function cameraError(error) {
+        console.debug("Unable to obtain picture: " + error, "app");
+
+    }, options);
+}
+```
+
+## Select an Image and Return Thumbnails (resized images) <a name="getFileThumbnails"></a>
+
+Resizing a file selected with the file picker works just like resizing using the Camera app. If you are using the file picker instead of the Camera, and you are resizing the image, the `Camera.EncodingType` value must match the value for the selected image. We previously set this value to JPEG in the app's `setOptions` function.
```
riknoll: I'm still a bit confused by the EncodingType comment. What platform has this behavior? Does specifying the encoding type filter what photos appear? I know it doesn't on Android, but I haven't done exhaustive testing on the other platforms and would be interested to know where you encountered that.

Mikejo5000: I spent some more time testing this; it is an issue that shows up on Windows 10 (x64/x86) but not on Android/iOS it appears. I think I uncovered a likely bug here; if you set config.xml to Win10 (and target Win10), the image URI returned by getPicture can't be used as an Image source (won't display when assigned to img.src); this specific issue described in the text only appears when config.xml is set to 8.1 AND you are targeting Win10. So the bigger issue appears to be a returned file URI on Win10 that is not easily used as an image source...

riknoll: Hmmm, that could be a bug. @rakatyal sound familiar? If you could give an example of the URI that is returned we can take a look.

riknoll: The other possibility is that the type of URI being returned just needs to be whitelisted as part of the content security policy in the HTML file. I've had CSP issues with some of the URIs returned on Android before

Mikejo5000: Yes, you're right! The issue is fixed when I add "ms-appdata:" to the CSP. A little tricky.

riknoll: Cool! That has confused me in the past as well and comes up on Slack sometimes too. Might be worth documenting!

riknoll:
LGTM! I'll merge this later today and regenerate the README

PR #205, opened 2016-04-13
CB-10873 Avoid crash due to usage of uninitialized variable when writ…

omefire:
…ing geolocation data to image destination

### DIFF
--- a/src/ios/CDVCamera.m
+++ b/src/ios/CDVCamera.m
@@ -359,22 +359,23 @@ static NSString* toBase64(NSData* data) {
                 data = UIImageJPEGRepresentation(image, 1.0);
             } else {
                 data = UIImageJPEGRepresentation(image, [options.quality floatValue] / 100.0f);
-                if (options.usesGeolocation) {
-                    NSDictionary* controllerMetadata = [info objectForKey:@"UIImagePickerControllerMediaMetadata"];
-                    if (controllerMetadata) {
-                        self.data = data;
-                        self.metadata = [[NSMutableDictionary alloc] init];
-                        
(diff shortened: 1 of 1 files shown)


omefire:
What happens is that when trying to write the image source, we do it with a null/nil variable ```self.data```:
https://github.com/omefire/cordova-plugin-camera/blob/CB-10873/src/ios/CDVCamera.m#L669-L671

omefire:
For more details, see: https://issues.apache.org/jira/browse/CB-10873

src/ios/CDVCamera.m
```diff
@@ -358,6 +358,7 @@ - (NSData*)processImage:(UIImage*)image info:(NSDictionary*)info options:(CDVPic
                 // use image unedited as requested , don't resize
                 data = UIImageJPEGRepresentation(image, 1.0);
             } else {
+                data = UIImageJPEGRepresentation(image, [options.quality floatValue] / 100.0f);
                 if (options.usesGeolocation) {
```
jasongin: When using the "unedited" image above, the geolocation option is ignored -- is that correct behavior? Or should the if (options.usesGeolocation) be moved to outside/after the else block?

nikhilkh:
@alsorokin Is it possible for our appium test cover this scenario? Sounds like you will have to enable this preference to see this issue.

omefire:
I agree on an appium test to cover this scenario, but I think it should go further than just making sure no crash happens and also validate that the EXIF header is correctly written: 
- https://github.com/gomfunkel/node-exif

I have the feeling that code path might not be well tested/exercised.

src/ios/CDVCamera.m
omefire: @jasongin I suspect it might not be correct behavior.

jasongin:
LGTM

PR #210, opened 2016-05-07
CB-10139 browser: Respect target width and height

kicken:
Fixes [CB-10139 Camera plugin is cropping image taked from camera in browser platform](https://issues.apache.org/jira/browse/CB-10139).

Cropping occurs due to the canvas's with and height not being set.  This patch sets the canvas width and height to the values specified in the options given to getPicture.

Tested in recent versions of Chrome, Firefox, and Edge and found no problems.

### DIFF
--- a/src/browser/CameraProxy.js
+++ b/src/browser/CameraProxy.js
@@ -23,7 +23,7 @@ var HIGHEST_POSSIBLE_Z_INDEX = 2147483647;
 
 function takePicture(success, error, opts) {
     if (opts && opts[2] === 1) {
-        capture(success, error);
+        capture(success, error, opts);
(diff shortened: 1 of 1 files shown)


nikhilkh:
@kicken These are great changes - I'm curious what are you using the browser platform's camera implementation for?

kicken:
@nikhilkh Nothing in particular.  My main target is a mobile app, but I happen to do a lot of preliminary testing in google chrome's device mode using the browser platform. Since I may eventually release a website to equivalent of the app, I figured I may as well try and re-use as much as I can so I occasionally try the app in normal browser mode and fix it up if it doesn't take too much effort.

PR #211, opened 2016-05-08
CB-11227 browser: Fix incorrect mime type

kicken:
Fixes [CB-11227: Incorrect type in toDataUrl call](https://issues.apache.org/jira/browse/CB-11227).


nikhilkh:
We'll need you to sign the ICLA here: http://www.apache.org/licenses/icla.txt and send it to secretary@apache.org to accept this PR.

PR #213, opened 2016-05-10
Add pull request template

nikhilkh:


riknoll:
Do we need to provide a link for CCLA info just in case? Can we also suggest a format for PR titles? Something like:
```
[CB-12345]: Short description
```

purplecabbage:
+1 to the PR title suggestion
The icla is NOT actually required for smaller changes, and we should only be enforcing this for large/deep changes.  (large,small,deep are all judgement calls I know)
The simple act of sending a pr does technically indicate that they are 'giving' the code to us, so we only need to protect ourselves for the larger changes.

tony--:
There are a couple places that contributor workflow is documented:
The wiki has a [contributor workflow page](https://wiki.apache.org/cordova/ContributorWorkflow) which also points to the [issue workflow page](https://wiki.apache.org/cordova/IssueWorkflow).
The docs have a [Contributor Guidelines page](http://cordova.apache.org/contribute/contribute_guidelines.html).

I guess the wiki pages are pretty much obsolete and the docs workflow is more current.

It seems like it would make sense to at least point to the docs page from the template, something like:
"This checklist is intended as a quick reference, for complete details please see our [Contributor Guidelines](http://cordova.apache.org/contribute/contribute_guidelines.html)".  
It would make sense to have the template checklist reflect the outline structure of the docs page workflow.  I guess this would mean add steps for getting code reviewed, running tests and commit message guideline.

If any requirements are added (e.g., PR title guideline) or changed as part of this, the docs should be updated to match the template checklist.

purplecabbage:
great points @tony-- ! +1 to all of 'em

nikhilkh:
Thanks for the feedback! I made some updates based on feedback. 

As far as running tests - we're automating that process using Jenkins, Travis CI as appropriate.

PR #243, opened 2016-12-16
Double exif rotation on android with camera

Webpal:
### Platforms affected
Android

### What does this PR do?
This solves the "correctOrientation not rotating / wrong orientation with Android camera bug"

Single Android device

If flag correctOrientation is set the image is rotated according to EXIF-data. After this, the original EXIF-data is added back to the image where the rotation is stil there. Result is that instead of a wrongly rotated image with EXIF-data to correct you end up with a correctly rotated image with EXIF-data that tilts the image 90 degrees.

Removing the EXIF completely solves this for me - there should be a more elegant solution to just remove the "orientation" data from the EXIF to make this a more general purpose solution.

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -556,13 +556,14 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
                 os.close();
 
                 // Restore exif data to file
+                   /*
                 if (this.encodingType == JPEG) {
                     String exifPath;
                     exifPath = uri.getPath();
                     exif.createOutFile(exifPath);
                     exif.writeExifData();
                 }
-
+                     */
                 // Send Uri back to JavaScript for viewing image
                 this.callbackContext.success(uri.toString());
 


infil00p:
What device does this happen to? We've seen reports of Samsung devices that have had that problem, and if we're already writing the EXIF with the rest of the data, there's no need to write EXIF twice.

Webpal:
Not sure if this should be merged - someone should look deeper into if this is the root cause of the issue and then apply a fix that re-applies EXIF but without rotation since the image has already been rotated... This solved the issue for my particular use case on a Galaxy S7 and the conclusion that the EXIF still keeps the "rotation 90 degrees" info when reapplied is just a theory that solved the issue for me. The final image should however probably stil have the EXIF data applied (just not the rotation part...)

sarantist:
Same problem exists for Samsung Galaxy S3, S4 and S5. All of the other test devices work as intended.

dantheman213:
I am having this problem and it is incredibly annoying :/

infil00p:
So, yeah, after testing this patch on a Google Pixel, I can't accept it as it is because we lose all our EXIF data if we remove this code.

infil00p:
OK, found the cause of the issue.  Just dumping EXIF isn't the answer.  I'm going to close this in favour of this PR: https://github.com/apache/cordova-plugin-camera/pull/252

PR #280, opened 2017-07-14
feat(browser): automatically click on file input field

MaxXx1313:
### Platforms affected
Browser

### What does this PR do?
Skip the step to load an image from file system when camera.getPicture() ought to load file from the file system (not take it from camera) 

I've run `npm test`. 
Actually, there is no test for browser platform. However these changes are used in a real project.


sandstrom:
This won't work in some browsers afaik, for example in Safari.

*Just a note, may be helpful to know.*

janpio:
Let's rerun this with current tests.

PR #289, opened 2017-10-06
CB-13332: (iOS) Error When Taking Picture Missing NSPhotoLibraryAddU…

mino922:
…sageDescription

### Platforms affected
iOS

### What does this PR do?
Fixes issues in iOS when taking a photo where app would crash because of missing NSPhotoLibraryUsageDescription

Change was tested on an ionic app on iPhone 6s running iOS 11.

### Checklist
- [X ] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [X] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [X] Added automated test coverage as appropriate for this change.

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -175,6 +175,11 @@
              <string>$PHOTOLIBRARY_USAGE_DESCRIPTION</string>
          </config-file>
 
+         <preference name="PHOTOLIBRARY_USAGE_DESCRIPTION" default=" " />
+         <config-file target="*-Info.plist" parent="NSPhotoLibraryUsageDescription">
+             <string>$PHOTOLIBRARY_USAGE_DESCRIPTION</string>
+         </config-file>
+
          <config-file target="*-Info.plist" parent="NSLocationWhenInUseUsageDescription">
              <string></string>
          </config-file>


mino922:
To use this, you must add to your config.xml 
`<variable name="PHOTOLIBRARY_ADD_USAGE_DESCRIPTION" value="Access Photo Library for Profile Picture" />`

jcesarmobile:
Thanks for you PR, but you are just duplicating NSPhotoLibraryUsageDescription, the one needed for iOS 11 is NSPhotoLibraryAddUsageDescription.

Anyway, we are removing the preferences like this from the plugin.xml as it causes conflicts with other plugins requiring the same permissions.
The correct way of adding the permissions is using `edit-config` tag in the config.xml like this

```
<edit-config target="NSPhotoLibraryAddUsageDescription" file="*-Info.plist" mode="merge">
    <string>need to save on the photo library for some reason</string>
</edit-config>
```

Can you close the PR?

marcovtwout:
FYI, `NSPhotoLibraryAddUsageDescription` is required when your plugin needs write access (shouldn't be required for read access) https://developer.apple.com/library/content/documentation/General/Reference/InfoPlistKeyReference/Articles/CocoaKeys.html

jcesarmobile:
Yeah, it's only required if you have `saveToPhotoAlbum` option set to `true`, for reading then you need `NSPhotoLibraryUsageDescription`

Either way, you have to use edit-config tag to set them, the plugin is removing the variables as they caused a lot of problems when other plugins wanted to set the same values

PR #294, opened 2017-10-20
fix for using the saveToPhotoAlbum option on iOS 11

joseph-perez:
iOS 11 added a new permission when trying to get write only access to
the photo library.

### Platforms affected
iOS 11

### What does this PR do?
Fixes the crash that occurs when trying to use the saveToPhotoAlbum option.

I tested my change on a iphone 7 running iOS 11.

### Checklist
- [x] [CB-13477](https://issues.apache.org/jira/browse/CB-13477) in the JIRA database

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -175,6 +175,11 @@
              <string>$PHOTOLIBRARY_USAGE_DESCRIPTION</string>
          </config-file>
 
+         <preference name="PHOTOLIBRARYADD_USAGE_DESCRIPTION" default=" " />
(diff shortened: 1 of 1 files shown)


jcesarmobile:
Thanks for the PR, but we are removing the variables from the plugins as they conflict with others.
The right way of setting the Usage Descriptions is edit-config tag.
See https://github.com/apache/cordova-plugin-camera/pull/295

Can you close the PR?

joseph-perez:
Yeah sure thing! Sorry about that

jcesarmobile:
no problems, thanks for contributing!

PR #295, opened 2017-10-20
CB-13264 (iOS): Remove ios usage descriptions

jcesarmobile:
### Platforms affected
iOS

### What does this PR do?
Removes the usage description variables and document how to set them using edit-config tab

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,23 +76,33 @@ Documentation consists of template and API docs produced from the plugin JS code
 
 ### iOS Quirks
 
-Since iOS 10 it's mandatory to add a `NSCameraUsageDescription` and `NSPhotoLibraryUsageDescription` in the info.plist.
+Since iOS 10 it's mandatory to provide an usage description in the `info.plist` if trying to access privacy-sensitive data. When the system prompts the user to allow access, this usage description string will displayed as part of the permission dialog box, but if you didn't provide the usage description, the app will crash before showing the dialog. Also, Apple will reject apps that access private data but don't provide an usage description.
 
-- `NSCameraUsageDescription` describes the reason that the app accesses the user’s camera.
+This plugins requires the following usage descriptions:
+
+- `NSCameraUsageDescription` describes the reason that the app accesses the user's camera.
 - `NSPhotoLibraryUsageDescription` describes the reason the app accesses the user's photo library. 
+- `NSLocationWhenInUseUsageDescription` describes the reason the app accesses the user's location. (Set it if you have `CameraUsesGeolocation` preference set to `true`)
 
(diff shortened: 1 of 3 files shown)


kleeb:
its not working in Phonegap Build

jcesarmobile:
This just removes the usage description variables from the plugin.xml, so not sure how this "is not working in Phonegap build".

Anyway, this is not the place to report your Phonegap build problems

kleeb:
thats true, but there is no way to add those params to the plist file with the latest update of PGB
which more or less makes the cordova-plugin-camera version stuck at 2.4.1...

unless using cordova-plugin-ios-camera-permissions

jcesarmobile:
Yes, there is a way, you have to use `edit-config` tag as documented in the README.md

kleeb:
edit-config doesn't work in the Phonegap Build version cli-7.1.0

kleeb:
just stating that thousands of PGB users will soon have problem when they update this plugin

jcesarmobile:
edit-config used to work in PGB. If it's now broken in cli-7.1.0 report it to Adobe, this plugin is from Apache.

jcesarmobile:
@kleeb FYI, I've created a sample app and built it in PGB setting the cli-7.1.0 and the texts were set correctly. You can try it yourself https://github.com/jcesarmobile/cordova-plugin-camera-sample.

First time I built it without version and was built with 6.5.0 and didn't include the texts but then I changed the CLI version, built it again and worked fine.

kleeb:
Thanks @jcesarmobile 
For this repo it works like charm

For my project it doesn't, some other plugin must be removing them.

kleeb:
Ok, we are using the old builder, and then it is not working

<preference name='pgb-builder-version' value='1' />

jcesarmobile:
If you want to use the old builder then you will need to use older versions of the plugins.
Anyway, if you have doubts or problems with Phonegap Build you should go to their forums https://forums.adobe.com/community/phonegap/build

kleeb:
I don't agree with that. This is just the different way how PGB is building the native package.
I don't see any connection with the plugin versions.

jcesarmobile:
To use latest version of the plugins you need to use `edit-config` tag, period.

If Phonegap Build old builder doesn't allow to use `edit-config` tag you have a few options:

1. Use the new builder.
2. Use an older version of the plugin where `edit-config` tag is not need.
3. Use another plugin that writes the usage descriptions.

Anyway, I have told you a few times that this plugins belongs to Apache and Phonegap Build belongs to Adobe, they have to adapt to our changes, not the other way around. And in this case they already provide a way, so it's you who has to switch to that or find a workaround.

Please, stop commenting here.

PR #320, opened 2018-04-27
Get applicationId from package name

squallstar:
I know this doesn't have a JIRA ticket assigned to it, so please feel free to link if there's already one issue (I couldn't see any).

### Platforms affected

- Android

### What does this PR do?

Makes sure the `applicationId` is set by reading from the app's package name instead of Gradle's BuildConfig file.

This is because apps can have a package name different from the manifest's package, but this results in the `BuildConfig` file being available under the mainfest's package, not the app's package name so it resulted in a failure when opening the BuildConfig class file.

I've been testing this extensively on Android using simulators and physical devices on multiple versions of the SDK.


janpio:
Hey, I just fixed the problem that caused Android tests to fail in `master`. Could you rebase this PR please? This should get rid of the Android failures and possibly fix all test failures for this PR.

squallstar:
@janpio done, can you check if it's correct now? This is what I've done:

```
git remote add edx https://github.com/apache/cordova-plugin-camera.git
git fetch edx
git rebase edx/master
git push -f
```

janpio:
Yep, merge commit looks clean: https://github.com/apache/cordova-plugin-camera/pull/320/commits/d1516bd0966608736e62423f18bc1619b02e2cea

We'll know in a bit if the tests succeed ;)

janpio:
Yes, Travis now happy 👍

squallstar:
Great! Feel free to merge anytime.

janpio:
(Just noticed the merge _is_ a bit messed up as it pulled in all the changes into your branch individually, but your actual change is simple enough:

```diff
-        this.applicationId = (String) BuildHelper.getBuildConfigValue(cordova.getActivity(), "APPLICATION_ID");
-        this.applicationId = preferences.getString("applicationId", this.applicationId);
+        this.applicationId = cordova.getContext().getPackageName();
```

I looked into the history of that code you removed:
Initially added here: https://github.com/squallstar/cordova-plugin-camera/commit/00e0a7dc469133faf92b46f0d72112d4d873844e#diff-f2e83abdaf34884df7e74b6493cf2dd7L134
The value used later in these 2 locations:
https://github.com/squallstar/cordova-plugin-camera/commit/00e0a7dc469133faf92b46f0d72112d4d873844e#diff-f2e83abdaf34884df7e74b6493cf2dd7R294
https://github.com/squallstar/cordova-plugin-camera/commit/00e0a7dc469133faf92b46f0d72112d4d873844e#diff-f2e83abdaf34884df7e74b6493cf2dd7L794

Then modified here: https://github.com/squallstar/cordova-plugin-camera/commit/61064ae3ed6c520e6de6646566b83772854d0fea

Correct that this value (plus `.provider` added at the end) is mainly used for getting a file name? I don't know anything about Android, is what and where you are getting the value from identical?

Mykaelos:
Yah, the `${applicationId}.provider` value is the authority used to allow the app access to that storage location. The authority has to match the one defined in the [authorities config](https://github.com/apache/cordova-plugin-camera/blob/master/plugin.xml#L65).

### Background
After a lot of investigation, I think I understand this now. The .BuildConfig class is an Android auto generated class that [BuildHelper.getBuildConfigValue](https://github.com/apache/cordova-plugin-compat/blob/master/src/android/BuildHelper.java#L50) is trying to access via reflection to find the APPLICATION_ID field.

Example BuildConfig.java that sits in build/generated/source/buildConfig/debug/com/company/project/
```
/**
 * Automatically generated file. DO NOT MODIFY
 */
package com.company.project;

public final class BuildConfig {
  public static final boolean DEBUG = Boolean.parseBoolean("true");
  public static final String APPLICATION_ID = "com.company.project";
  public static final String BUILD_TYPE = "debug";
  public static final String FLAVOR = "";
  public static final int VERSION_CODE = 1;
  public static final String VERSION_NAME = "2018.1-SNAPSHOT";
}
```

We were running into the same issue where BuildHelper.getBuildConfigValue was returning null and crashing the app, because we didn't have the `<preference name="applicationId" value="{PACKAGE_ID}"/>` set either. This is because the application's package name is different from the one that's ultimately built into the APK. We updated the package name of the app, because it's a new version of the app with new branding, but we need to submit the previous package name in the APK so it can goto the same Google Store page.

build.gradle
```
android {
    defaultConfig {
        applicationId "com.company.oldname"
    }
}
```
`Java app package: com.company.newname;`

In our example, `cordova.getActivity().getPackageName() == "com.company.oldname"` but the `com.company.oldname.BuildConfig` doesn't exist (Android has created the `com.company.newname.BuildConfig`). Because `com.company.oldname.BuildConfig` doesn't exist, BuildHelper.getBuildConfigValue returns null. In most cases, BuildHelper.getBuildConfigValue works fine, but this is an edge case where this fails. This would also fail for apps with [multiple productFlavors](http://tools.android.com/tech-docs/new-build-system/applicationid-vs-packagename).

### Possible Solution
The Camera Plugin could support all three options: 
- First use preference.applicationId. 
- If null, use BuildHelper.getBuildConfigValue.
- If null, use cordova.getActivity().getPackageName() and hope it's good enough.

I'm not sure if `cordova.getActivity().getPackageName()` will always be the right package name because I'm not sure if the Cordova activity/context would have the same package name as the app using it. But I think it's worth it to try to prevent this.applicationId from just being null to crash later on.

Important note, CordovaInterface.getContext() was added in Cordova Android 7.0, so it might be better to use cordova.getActivity() to support previous versions (this is important to us because we're stuck with Cordova Android 6.3.0 for the minute).

Let me know if you want me to submit a PR with these suggested changes, or if you want to just make them in this PR.

jcesarmobile:
This is fixed in cordova-android 9 and the next release of the plugin will require cordova-android 9, so it's not needed anymore.

PR #321, opened 2018-05-08
CB-14082: (iOS) Dismiss camera popover on permission denied

KamenGoranchev:
### Platforms affected
IOS

### What does this PR do?
It dismisses the camera popover automatically when the permission for camera usage is denied

Used the patched version of the plugin in a iOS application going through all the cases of accepting/denying permissions

### Checklist

### DIFF
--- a/src/ios/CDVCamera.m
+++ b/src/ios/CDVCamera.m
@@ -179,6 +179,22 @@ static NSString* toBase64(NSData* data) {
                                       cancelButtonTitle:NSLocalizedString(@"OK", nil)
                                       otherButtonTitles:settingsButton, nil] show];
                 });
+            } else if (authStatus == AVAuthorizationStatusNotDetermined){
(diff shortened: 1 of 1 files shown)


wtrocki:
Quick question:
New code added will just pop up dialog when permission was not granted/rejected before and then we manually request permission and interact with that. 

Would it be better to also support this when `Permission is Denied` before (see code above your change)?

kohlia:
I'd like this to be included as I am using the plugin and on iOS a black image is captured if permission is denied

KamenGoranchev:
@wtrocki Sorry I don't think I understand your question.

The code above my changes is handling the situation correctly. It does show a custom alert and closes the camera popover when the user dismisses the alert. I wanted to simulate the same behavior also in case with the system camera permission alert. The problem there was that I could not register callback to the alert that pops up automatically , therefor I needed to request the permission manually.

So if your question is why we don't do the same thing in the other case the answer is that it is already done there. Or if ask why we don't ask for the permission one more time when it is already denied the answer is that iOS does not support asking for the same permission if it is once denied.

PR #324, opened 2018-05-31
CB-14113: document issue workflow

niccokunzmann:
### What does this PR do?
This documents the issue workflow for me as a first-timer.
- refer to issue documentation
- name the project to file the issue under

Reading the docs twice.

### Checklist
- [X] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [X] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [X] Added automated test coverage as appropriate for this change.


jcesarmobile:
Thanks for the PR, but we are planning on moving to github issues soon, so this will become irrelevant. 
Also doing it for just one repo makes no sense, in case we do it, we should do it for all plugins and platforms, but as I said, there is no point as we are going to move to github issues.

PR #325, opened 2018-06-06
Got an activity result, but no plugin was registered to receive it.

OSpoon:
android 6.0.1
cordova 8.0.0
cordova-plugin-camera 4.0.3

After taking photos, it will repeat the output of the title of the title, resulting in the failure to get the image.


OSpoon:
The reason is that onActivityResult is rewritten in MainActivity in my Cordova project, and Android permissions are processed in onActivityResult.

PR #339, opened 2018-09-10
Localizing hardcoded messages

kohlia:
### Platforms affected
iOS

### What does this PR do?
Localizes messages hardcoded

Works on Cordova based application 

### Checklist
- [ ] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -100,6 +100,7 @@
             <clobbers target="CameraPopoverHandle" />
          </js-module>
 
+         <resource-file src="src/ios/Camera.strings" target="Resources/en.lproj/Camera.strings" />
          <header-file src="src/ios/UIImage+CropScaleOrientation.h" />
          <source-file src="src/ios/UIImage+CropScaleOrientation.m" />
          <header-file src="src/ios/CDVCamera.h" />
--- a/plugin.xml
+++ b/plugin.xml
@@ -100,7 +100,7 @@
             <clobbers target="CameraPopoverHandle" />
          </js-module>
 
-         <resource-file src="src/ios/Camera.strings" target="Resources/en.lproj/Camera.strings" />
+         <resource-file src="src/ios/Camera.strings" target="en.lproj/Camera.strings" />
          <header-file src="src/ios/UIImage+CropScaleOrientation.h" />
          <source-file src="src/ios/UIImage+CropScaleOrientation.m" />
          <header-file src="src/ios/CDVCamera.h" />
--- a/src/ios/Camera.strings
+++ b/src/ios/Camera.strings
@@ -1,0 +1,1 @@
+"camera.prohibited" = "Access to the camera has been prohibited; please enable it in the Settings app to continue.";
--- a/src/ios/Camera.strings
+++ b/src/ios/Camera.strings
@@ -1,1 +1,3 @@
 "camera.prohibited" = "Access to the camera has been prohibited; please enable it in the Settings app to continue.";
+"camera.ok" = "Settings";
+"camera.settings" = "OK";
--- a/src/ios/CDVCamera.m
+++ b/src/ios/CDVCamera.m
@@ -166,7 +166,7 @@ static NSString* toBase64(NSData* data) {
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wtautological-pointer-compare"
                 NSString* settingsButton = (&UIApplicationOpenSettingsURLString != NULL)
-                    ? NSLocalizedString(@"Settings", nil)
+                    ? NSLocalizedStringFromTable(@"camera.settings", @"Camera", nil)
                     : nil;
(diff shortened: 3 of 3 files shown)


kohlia:
Please merge this

plugin.xml
```diff
@@ -100,6 +100,7 @@
             <clobbers target="CameraPopoverHandle" />
          </js-module>
 
+         <resource-file src="src/ios/Camera.strings" target="en.lproj/Camera.strings" />
```
janpio: I don't know anything about iOS. What does `target="en.lproj/Camera.strings"` mean here?

kohlia: It will copy the hardcoded message into a language file named Camera.strings and we can load the message from this file

src/ios/Camera.strings
```diff
@@ -0,0 +1,3 @@
+"camera.prohibited" = "Access to the camera has been prohibited; please enable it in the Settings app to continue.";
+"camera.ok" = "Settings";
+"camera.settings" = "OK";
```
dpogue: These last two strings seem to be swapped

kohlia: Oh sorry about that should I update ?

plugin.xml
janpio: Ok, so `en.lproj` is a magic thing that the project will know about and load `en.lproj/Camera.strings` without any additional configuration?

Should the file maybe be called `cordova-plugin-camera.strings`?

kohlia: The en.lproj or fr.lproj or ar.lproj etc are understood by app to load corresponding language files
You are free to name the strings file just update the plugin.xml accordingly

janpio: That was really more a question - to make sure developers know what the file is from and for. You created the PR, so you can decide and adapt or not.

src/ios/Camera.strings
janpio: Yes.

kohlia: Done

oliversalzburg:
We use https://github.com/fairmanager-cordova/plugin-localization-strings for this. It applies pretty much universally to all native localizable strings. I like a centralized solution for this problem. Might be worth a look.

kohlia:
> We use https://github.com/fairmanager-cordova/plugin-localization-strings for this. It applies pretty much universally to all native localizable strings. I like a centralized solution for this problem. Might be worth a look.

I gather the values in InfoPlist.strings cannot be accessed in code, correct me if I am wrong

oliversalzburg:
@kohlia I'm not sure I understand the question correctly. The plugin will create localization projects from the JSON configuration. The remaining process is pretty much transparent.

If you need a localized string from ObjC, this is how to do it: https://github.com/fairmanager-cordova/plugin-barcodescanner/commit/73d0f497b242627b33dcd6fbfb32d0d14501b100

kohlia:
@oliversalzburg makes sense so now I need to update the code to fetch from InfoPlist.strings but this will add dependency on that plugin

oliversalzburg:
@kohlia I'm sorry, I didn't mean to suggest to depend on that plugin. I wanted to merely point out an existing solution in this area. This patch looks like it tries to achieve similar things and this approach might even conflict with existing solutions.

Ideally, all solutions play nice together. Sadly, I can't fully evaluate the situation at the moment.

kohlia:
So what now?

kohlia:
> Now you can decide if this has any impact on your PR and change something or not.
> And then some maintainer will maybe review your PR and decide if it can and should be merged or not.

This still does not impact the PR as dependency on another plugin should not be created. I feel this solution works fine

janpio:
Nobody told you to include the other plugin, @oliversalzburg  just mentioned it so we all can consider if these changes might make this plugin incompatible to that other plugin that solves a similar problem for existing users.

kohlia:
Its up to the maintainers now.

kohlia:
Please merge the change

janpio:
Hey, I just fixed the problem that caused Android tests to fail in `master`. Could you rebase this PR please? This should get rid of the Android failures and possibly fix all test failures for this PR.

jcesarmobile:
Strings are already localizable as they are `NSLocalizedString`, I don't see any improvement on changing them to `NSLocalizedStringFromTable`

kohlia:
That is done to specify language file, so as to keep if separate from overlapping any application language files

oliversalzburg:
I'd prefer if the literals were prefixed (as they already are in this patch) and just be left in the main table. Creating multiple tables might be the cleaner approach, but it's also a new solution in the entire environment and might require more thought in general.

guirip:
Hello there :wave: 
Any chance to see the conflict resolved? We need localization for this message but it doesn't seem possible today.

Issue #405, opened 2019-01-09
Support for Android 3.x and Jetpack


alsocalledchris:
Building a Cordova App that uses Android 3.x and Jetpack does not build with the below errors:

- Error:(25, 34) error: package android.support.v4.content does not exist

I'm not able to change the compile target version (28) which looks like a solution that others have made. Any other suggestions?

Thanks.

dfattlar:
+1  Seeing this issue too.

pbergner:
seeing it well, is there a workaround?

stck:
In order to support jetpack's format, dependency
`android.support.v4.content` should be changed to `androidx.loader.content`

dfattlar:
Thanks @stck! From Android Studio I clicked Refactor > Migrate to AndroidX. After the scan, I clicked 'Do Refactor', and was then able to build to Android device again.

contfedorov:
`cordova-android` supports AndroidX since [version 9.0](https://cordova.apache.org/announcements/2020/06/29/cordova-android-9.0.0.html). Are there any plans to move this plugin to AndroidX instead of support library?

jcesarmobile:
Version 6 of the plugin uses AndroidX 
https://github.com/apache/cordova-plugin-camera/pull/751

PR #418, opened 2019-01-30
refactor: Migrate to AndroidX

davidschreiber:
### Platforms affected

Android

### What does this PR do?

Migrate to AndroidX (resolves parts of https://github.com/apache/cordova-plugin-camera/issues/405). This plugin used a single class from the (now) deprecated Android Support Library:

* The dependency to the `com.android.support:support-v4` artifact was replaced with the newer `androidx.legacy:legacy-support-v4` as suggested by the AndroidX migration guides: https://developer.android.com/jetpack/androidx/migrate
* All use of `android.support.v4.content.FileProvider` was replaced with then newer `androidx.core.content.FileProvider` as suggested by the AndroidX migration guide.
* The plugin property `ANDROID_SUPPORT_V4_VERSION` was replaced with a more accurately named one called `ANDROIDX_LEGACY_SUPPORT_V4_VERSION` which now defaults to `1.0.0`.

Manual testing has been performed as part of reporting to the initial issue that lead to this PR: https://github.com/apache/cordova-android/issues/565

### Checklist
- [ ] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database


janpio:
Android builds on Travis are not happy. Can you take a look at those failures to determine if this can be caused by your changes?

brodybits:
> Android builds on Travis are not happy.

I suspect this is due to using old build tools and maybe old SDK in `.travis.yml`.

davidschreiber:
Here's an excerpt of the build errors on Travis:

```
:app:processDebugResources/home/travis/.gradle/caches/transforms-1/files-1.1/core-1.0.0.aar/b9b8c67c94d012f8e982861787cf3aec/res/values/values.xml:133:5-70: AAPT: error: resource android:attr/fontVariationSettings not found.
    
/home/travis/.gradle/caches/transforms-1/files-1.1/core-1.0.0.aar/b9b8c67c94d012f8e982861787cf3aec/res/values/values.xml:133:5-70: AAPT: error: resource android:attr/ttcIndex not found.
    
/tmp/tmp-38189QsOugRLFnQO/platforms/android/app/build/intermediates/incremental/mergeDebugResources/merged.dir/values/values.xml:87: error: resource android:attr/fontVariationSettings not found.
/tmp/tmp-38189QsOugRLFnQO/platforms/android/app/build/intermediates/incremental/mergeDebugResources/merged.dir/values/values.xml:87: error: resource android:attr/ttcIndex not found.
error: failed linking references.
```

The missing attributes `android:attr/fontVariationSettings` and `android:attr/ttcIndex` are used within the AndroidX library, and are part of the Android framework since API 28: https://developer.android.com/sdk/api_diff/28/changes/android.R.attr

This means that AndroidX does actually require `compileSdkVersion 28` in the build. This is also mentioned in the AndroidX docs (https://developer.android.com/jetpack/androidx/#using_androidx):
> If you want to use AndroidX in a new project, you need to set the compile SDK to Android 9.0 (API level 28) or higher and set both of the following Android Gradle plugin flags to true in your gradle.properties file.

I don't know my way around how Cordova builds and tests using Travis, so my first attempt was to simply add another target environment with API 28. This failed, since it seems that `cordova-paramedic` does not yet contain a configuration for that API level. I went ahead and forked `cordova-paramedic` and prepared an experimental branch where I added an Android 9.0 configuration: https://github.com/davidschreiber/cordova-paramedic/tree/davidschreiber/android-9

Let's see if Travis likes that. If I'm doing anything completely wrong, please let me know. Thanks!

davidschreiber:
Last build on travis failed because my `cordova-paramedic` configuration was still using the latest stable version of `cordova-android` which does not yet compile with API 28. I've not pushed another config that tries using a nightly. Can't tell right away if this works, but as far as I can tell the latest `cordova-android` uses `compileSdkVersion 28` by default: https://github.com/apache/cordova-android/blob/2163c8ae0536bfde9029cfc981b945f4af93e09f/bin/templates/project/build.gradle#L46

Trial and error.

davidschreiber:
Just found this issue https://github.com/apache/cordova-android/issues/631 which will probably cause the build to fail since it's using API 27.

brodybits:
Thanks @davidschreiber for all of your efforts. I think we should definitely remove Node.js 4 from .travis.yml and resolve apache/cordova-android#631 (default compile SDK version / default target SDK version issue) before you continue working on this one.

Note that we still need to continue testing on Node.js back to version 6 right now.

brodybits:
@davidschreiber are you still interested in getting this fixed?

I personally do not have so much time to work on it.

I would personally recommend that you raise each of the changes needed in separate PRs for us to review.

I would personally favor that we start a new major version that drops Node.js 6 and maybe even Node.js 8 to avoid the build problems.

janpio:
Well, a few months later the CI problems are fixed and this successfully passes. Could you have another look @davidschreiber if the changes make sense? How would one properly test this manually?

davidschreiber:
This is great @janpio! Thanks a lot for the update. Work on this PR has long been overdue on my end 🙈 

The simplest way to test is, probably is to create a sample project and use `cordova-plugin-androidx` to enable AndroidX and Jetifier inside the project. 

Adding the currently stable version of `cordova-plugin-camera` will yield an error after AndroidX was enabled inside the project (in this specific case several classes can't be found because the Support Library dependencies are no longer present since Gradle replaces them with AndroidX):

The easiest way to manually test the plugin is using the projects `tests/` sub projects. You can quickly create a test application using this:

```shell
# Clone the project, since cordova CLI does not support plugins from sub directories.
git clone https://github.com/davidschreiber/cordova-plugin-camera.git && cd cordova-plugin-camera && git checkout davidschreiber/androidx-migration && cd ..
# Create a test project.
cordova create androidx-camera-test com.example.cordovacameratest androidx-camera-test
cd androidx-camera-test
# Add Android platform support
cordova platform add android
# Add the test dependencies
cordova plugin add cordova-plugin-test-framework
cordova plugin add ../cordova-plugin-camera/tests
# Add the plugin itself
cordova plugin add ../cordova-plugin-camera
# Enable AndroidX inside the project
cordova plugin add cordova-plugin-androidx
# Set the entrypoint of the app to start the tests.
sed -i '' 's/index.html/cdvtests\/index.html/g' config.xml
# Build and run
cordova run android
```

You should now see the Jasmine test framework and can run the automated and manual tests of the plugin and verify that stuff behaves as it should.

Alternatively, if you want to test the current stable version of the plugin, you can run it like so:

```shell
git clone https://github.com/apache/cordova-plugin-camera.git && cd cordova-plugin-camera && git checkout 4.1.0 && cd ..
cordova create androidx-camera-test com.example.cordovacameratest androidx-camera-test
cd androidx-camera-test
cordova platform add android
cordova plugin add cordova-plugin-test-framework
cordova plugin add ../cordova-plugin-camera/tests
cordova plugin add ../cordova-plugin-camera
cordova plugin add cordova-plugin-androidx
sed -i '' 's/index.html/cdvtests\/index.html/g' config.xml
cordova run android
```

This will yield a build error due to the sources of the stable version still referencing old Support Library classes

```
...
> Task :app:compileDebugJavaWithJavac FAILED
/Users/davidschreiber-ranner/temp/androidx-update-test/platforms/android/app/src/main/java/org/apache/cordova/camera/CameraLauncher.java:64: error: package android.support.v4.content does not exist
import android.support.v4.content.FileProvider;
...
```

I hope this helps to verify that the changes in this PR are good to go. Please let me know if I can do anything else here.

janpio:
No, perfect explanation - will come back in a bit with my results.

(Ping @dpa99c for a check here, as this uses your plugin!)

(Semi offtopic: Should we add a test run that adds `cordova plugin add cordova-plugin-androidx` to all our plugins?)

davidschreiber:
Don't worry, we're performing tests for our own Cordova plugin like this too, so I already knew the procedure.

Adding an extra test for AndroidX support would definitely be a good step in the right direction — AndroidX is going to stay in the Android ecosystem.

janpio:
🎁 Paramedic feature request being tracked in https://github.com/apache/cordova-paramedic/issues/189



> Don't worry, we're performing tests for our own Cordova plugin like this too, so I already knew the procedure.

CI or manually? Public? Couldn't find any obvious CI config on the 2 Cordova plugin repos I know from you.

davidschreiber:
Manually and private. We haven't found any satisfying solution to automate tests for our plugin. We'd need some testing framework that invokes our plugin's APIs in JavaScript and performs assertions in Java and JavaScript (something we haven't figured out yet).

janpio:
Contacted you via LinkedIn with my email address, so we can talk about that further.

---

Is it correct that using this plugin in an AndroidX enabled app would currently crash with these during build:

```
/tmp/tmp-3971dYpBCp3HDhgA/platforms/android/app/src/main/java/org/apache/cordova/camera/CameraLauncher.java:40: error: package android.support.v4.content does not exist
import android.support.v4.content.FileProvider;
                                 ^
/tmp/tmp-3971dYpBCp3HDhgA/platforms/android/app/src/main/java/org/apache/cordova/camera/CordovaUri.java:25: error: package android.support.v4.content does not exist
import android.support.v4.content.FileProvider;
                                 ^
/tmp/tmp-3971dYpBCp3HDhgA/platforms/android/app/src/main/java/org/apache/cordova/camera/FileProvider.java:21: error: package android.support.v4.content does not exist
public class FileProvider extends android.support.v4.content.FileProvider {}
                                                            ^
/tmp/tmp-3971dYpBCp3HDhgA/platforms/android/app/src/main/java/org/apache/cordova/camera/CameraLauncher.java:303: error: cannot find symbol
        this.imageUri = new CordovaUri(FileProvider.getUriForFile(cordova.getActivity(),
                                                   ^
  symbol:   method getUriForFile(Activity,String,File)
  location: class FileProvider
/tmp/tmp-3971dYpBCp3HDhgA/platforms/android/app/src/main/java/org/apache/cordova/camera/CameraLauncher.java:801: error: cannot find symbol
                        Uri tmpFile = FileProvider.getUriForFile(cordova.getActivity(),
                                                  ^
  symbol:   method getUriForFile(Activity,String,File)
  location: class FileProvider
```

And that after merging these changes, that would not be the case any more?

---

@brodybits This should be considered a BREAKING change because we are taking `android.support.v4.content` away and replacing it with Androidx equivalents?

brodybits:
Yes I would consider this to be a breaking change.

davidschreiber:
Yes, update to AndroidX should be considered a breaking change. You are also correct about the symbol errors you are seeing when using the stable version of this plugin in an AndroidX-enabled app. 

I'm a bit torn about the idea of adding `cordova-plugin-androidx` (or [`cordova-plugin-androidx-adapter`](https://github.com/dpa99c/cordova-plugin-androidx-adapter)) as a dependency to this plugin. Doing so would impose a "single right way" to implement AndroidX in your Cordova application — which in my opinion does not exist. For example, developers could also manually enable AndroidX in their app or use different plugins that do the job via hooks. 

On the other hand, from my experience, I can tell that most Cordova users wouldn't be able to identify AndroidX/Support Library conflicts in their apps, let alone know how to resolve them. While adding documentation would be a good first step, making sure they never run into these kinds of issues (i.e. by automatically transitioning their app to AndroidX using the aforementioned plugins) would definitely be helpful for most users.

If the plugin should depend on a specific plugins that enable AndroidX, I would argue that they should be official Apache Cordova plugins. (In other words, the AndroidX support should be provided by the framework itself.)

janpio:
So we should probably move this discussion to `cordova-android` and find a more general solution?

janpio:
As I just stumbled over this: https://github.com/EddyVerbruggen/nativescript-plugin-firebase/pull/1261/files On NativeScript plugins they seem to have a way to differentiate if they need AndroidX support or not, which makes it possible for the plugins to be used in both scenarios. MAybe this is possible here as well somehow?

PR #435, opened 2019-03-29
Update browser plugin

GedasGa:
### Platforms affected

Browser and Electron

### Motivation and Context

Using the camera plugin with the `Browser` platform you can not specify the location, where the camera opens. Once you call `navigator.camera.getPicture`, for example, from your `index.js` with `sourceType: Camera.PictureSourceType.CAMERA` option, it creates a container and opens the Camera in it. 

The problem is, the container with the video is always appended to the `document.body` and the user has no control of it, except the `width` and `height`. The only way to customize its style is by selecting `cordova-camera-capture` class and applying your styles. Besides, this makes close impossible to open the camera inside a modal, unless you're willing to do crazy DOM manipulation after you call  `navigator.camera.getPicture` method. Same applies for the capture button.

Moreover, I'm missing the `Cancel/Close` button because once you open the camera, now there's no way to close it unless you take a picture.

Similarly when you call `navigator.camera.getPicture`, for example, from your `index.js` with `sourceType: Camera.PictureSourceType.PHOTOLIBRARY ` or `sourceType: Camera.PictureSourceType. SAVEDPHOTOALBUM ` option, it creates the input field.This field is also appended to the `document.body` and the user has no control of it.

I've felt, the need to allow the user to choose where he wants to open the camera, which `capture`/ `cancel` button or which input field he wants to use. For example, the user could create a container element and provide us with the ID of the element, where he wants to open the camera. This would give the user more control of the camera plugin and let easier modify its container. This would allow the user to easily open the Camera in a modal or where ever he wants.

### Description

Allow users to set *4* additional `options`:
    - `customCameraContainer` _(id of custom camera's container element)_
    - `customCaptureButton` _(id of custom camera's capture button)_
    - `customCancelButton` _(id of custom camera's cancel button)_
    - `customSourceInput` _(id of the custom source input element, when using different mode than the default sourceType)_

The default values of these options are `null`. If the user does not provide the `ID` of a *custom element*, the *default element will be used.

Besides, these changes, I've refactored the code to use `const` and `let` instead of `var`. Also, I've broken down big function to functions smaller.

### Testing

`npm t`

I've already tested the code manually.

Also tested the following cases manually.

- **Calling `navigator.camera.getPicture` with *no options*:**

- **Results:**
Once function is called, it create the following container. Streams a video form the Camera. 
    - If `Capture` is clicked, it returns the image data as `base64`, stops video stream and deletes the `<div>` element.
    - If `Cancel`is clicked, it stops video stream and deletes the `<div>` element.
```
<div class="cordova-camera-capture" style="position: relative; z-index: 2147483647;">
    <video width="320" height="240" src="blob:file:///90647c34-f521-4c8d-9df4-1224f60c981e"></video>
    <button>Capture</button>
    <button>Cancel</button>
</div>
```
------------------------------------------
- **Calling `navigator.camera.getPicture` with *options*:**
```
    sourceType: Camera.PictureSourceType.PHOTOLIBRARY
```

- **Results:**
Once function is called, it create the following input field.
    - If `Input` is clicked, it open default file manager and allows to upload 1 image. Return the image data as a callback.
```
<input class="cordova-camera-select" type="file" name="files[]" style="position: relative; z-index: 2147483647;">
```
------------------------------------------
- **Calling `navigator.camera.getPicture` with *options*:**
```
    sourceType: Camera.PictureSourceType.PHOTOLIBRARY,
    customCancelButton: 'cordova-source-input'
    
```

- **Results:**
Once function is called, it create the following input field.
    - If `Input` is clicked, it open default file manager and allows to upload 1 image. Return the image data as a callback.
```
<input class="cordova-source-input" type="file" name="files[]">
```

------------------------------------------
   - **Calling `navigator.camera.getPicture` with *options*:**
   ```
       customCameraContainer: 'cordova-camera-preview',
   ```
   
   - **Results:**
   Once function is called, it create the following container. Streams a video form the Camera. 
       - If `Capture` is clicked, it returns the image data as `base64`, stops video stream and deletes everything inside `<div>` element.
       - If `Cancel`is clicked, it stops video stream and deletes the `<div>` everything inside  element.
   ```
   <div id="cordova-camera-preview">
       <video width="320" height="240" src="blob:file:///90647c34-f521-4c8d-9df4-1224f60c981e"></video>
       <button>Capture</button>
       <button>Cancel</button>
   </div>
   ```

------------------------------------------
- **Calling `navigator.camera.getPicture` with *options*:**
```
    customCameraContainer: 'cordova-camera-preview',
    customCaptureButton: 'cordova-camera-capture',
    customCancelButton: 'cordova-camera-cancel'
```

- **Results:**
Once function is called, it create the following container. Streams a video form the Camera. 
    - If `Capture` is clicked, it returns the image data as `base64`, stops video stream and deletes the `<video>` element.
    - If `Cancel`is clicked, it stops video stream and deletes the `<video>` element.
```
<div id="cordova-camera-preview">
    <video width="320" height="240" src="blob:file:///90647c34-f521-4c8d-9df4-1224f60c981e"></video>
    <button id="cordova-camera-capture">Capture</button>
    <button  id="cordova-camera-cancel">Cancel</button>
</div>
```

### Checklist


README.md
```diff
@@ -266,6 +266,10 @@ Optional parameters to customize the camera settings.
 | saveToPhotoAlbum | <code>Boolean</code> |  | Save the image to the photo album on the device after capture. |
 | popoverOptions | <code>[CameraPopoverOptions](#module_CameraPopoverOptions)</code> |  | iOS-only options that specify popover location in iPad. |
 | cameraDirection | <code>[Direction](#module_Camera.Direction)</code> | <code>BACK</code> | Choose the camera to use (front- or back-facing). |
+| customCameraContainer | <code>string</code> |  | Browser-only option, specify the id of custom camera's container element.  |
```
erisu: [1] IMO, I think it would be better if these new options accept either a `query` string or `DOMElement`.

* `Query` string would give more flexibility in how the element is discovered vs an ID only. In this case, we would need to use the `document.querySelector` method instead of the `getElementById`.

* `DOMElement` could also be a viable option.

README.md
```diff
@@ -266,6 +266,10 @@ Optional parameters to customize the camera settings.
...
+| customCaptureButton | <code>string</code> |  | Browser-only option, specify the id of custom camera's capture button. |
```
erisu: Same as above [1].

README.md
```diff
@@ -266,6 +266,10 @@ Optional parameters to customize the camera settings.
...
+| customCancelButton | <code>string</code> |  | Browser-only option, specify the id of custom camera's cancel button.  |
```
erisu: Same as above [1].

README.md
```diff
@@ -266,6 +266,10 @@ Optional parameters to customize the camera settings.
...
+| customSourceInput | <code>string</code> |  | Browser-only option, specify the id of custom  source input element, when using diffrerent mode than the default `sourceType`. |
```
erisu: Same as above [1].

src/browser/CameraProxy.js
```diff
@@ -19,103 +19,187 @@
...
+    } else if (parent.hasChildNodes() && !customElements.customCancelButton) {
+        parent.removeChild(video);
+        parent.removeChild(parent.lastChild);
+    } else {
+        parent.removeChild(video);
+    }
+}
+
+function createSourceInput () {
+    let input = document.createElement('input');
```
erisu:
```suggestion
    const input = document.createElement('input');
```

src/browser/CameraProxy.js
```diff
@@ -19,103 +19,187 @@
...
+    input.className = 'cordova-camera-select';
+    input.type = 'file';
+    input.name = 'files[]';
+
+    return input;
+}
+
+function handleSourceInput (successCallback, sourceInput) {
+    sourceInput.onchange = function (inputEvent) {
+        let reader = new FileReader(); /* eslint no-undef : 0 */
```
erisu:
```suggestion
        const reader = new FileReader(); /* eslint no-undef : 0 */
```

src/browser/CameraProxy.js
```diff
@@ -19,103 +19,187 @@
...
+function capture (successCallback, errorCallback, opts) {
+    let targetWidth = opts[3];
+    let targetHeight = opts[4];
+    const customCameraContainer = opts[12];
+    const customCaptureButton = opts[13];
+    const customCancelButton = opts[14];
+
+    const customElements = {customCameraContainer, customCaptureButton, customCancelButton};
+
+    let parent = customCameraContainer ? document.getElementById(customCameraContainer) : createCameraContainer();
```
erisu:
```suggestion
    const parent = customCameraContainer ? document.getElementById(customCameraContainer) : createCameraContainer();
```

src/browser/CameraProxy.js
```diff
@@ -19,103 +19,187 @@
...
+    let video = createVideoStreamContainer(parent, targetWidth, targetHeight);
```
erisu:
```suggestion
    const video = createVideoStreamContainer(parent, targetWidth, targetHeight);
```

src/browser/CameraProxy.js
```diff
@@ -19,103 +19,187 @@
...
+    let captureButton = customCaptureButton ? document.getElementById(customCaptureButton) : createButton(parent, 'Capture');
```
erisu:
```suggestion
    const captureButton = customCaptureButton ? document.getElementById(customCaptureButton) : createButton(parent, 'Capture');
```

src/browser/CameraProxy.js
```diff
@@ -19,103 +19,187 @@
...
+    let cancelButton = customCancelButton ? document.getElementById(customCancelButton) : createButton(parent, 'Cancel');
```
erisu:
```suggestion
    const cancelButton = customCancelButton ? document.getElementById(customCancelButton) : createButton(parent, 'Cancel');
```

src/browser/CameraProxy.js
```diff
@@ -19,103 +19,187 @@
...
+    handleCaptureButton(successCallback, errorCallback, captureButton, video, customElements);
+    handleCancelButton(cancelButton, video, customElements);
 }
 
-function capture (success, errorCallback, opts) {
-    var localMediaStream;
-    var targetWidth = opts[3];
-    var targetHeight = opts[4];
+function createCameraContainer () {
+    let parent = document.createElement('div');
```
erisu:
```suggestion
    const parent = document.createElement('div');
```

src/browser/CameraProxy.js
```diff
@@ -19,103 +19,187 @@
...
     targetWidth = targetWidth === -1 ? 320 : targetWidth;
     targetHeight = targetHeight === -1 ? 240 : targetHeight;
 
-    var video = document.createElement('video');
-    var button = document.createElement('button');
-    var parent = document.createElement('div');
-    parent.style.position = 'relative';
-    parent.style.zIndex = HIGHEST_POSSIBLE_Z_INDEX;
-    parent.className = 'cordova-camera-capture';
+    let video = document.createElement('video');
```
erisu:
```suggestion
    const video = document.createElement('video');
```

src/browser/CameraProxy.js
```diff
@@ -19,103 +19,187 @@
...
+    video.width = targetWidth;
+    video.height = targetHeight;
+
     parent.appendChild(video);
+
+    return video;
+}
+
+function createButton (parent, innerText) {
+    let button = document.createElement('button');
```
erisu:
```suggestion
    const button = document.createElement('button');
```

src/browser/CameraProxy.js
```diff
@@ -19,103 +19,187 @@
...
 
-    button.onclick = function () {
+function handleCaptureButton (successCallback, errorCallback, captureButton, video, customElements) {
+    captureButton.onclick = function () {
         // create a canvas and capture a frame from video stream
-        var canvas = document.createElement('canvas');
-        canvas.width = targetWidth;
-        canvas.height = targetHeight;
-        canvas.getContext('2d').drawImage(video, 0, 0, targetWidth, targetHeight);
+        let canvas = document.createElement('canvas');
```
erisu:
```suggestion
        const canvas = document.createElement('canvas');
```

src/browser/CameraProxy.js
```diff
@@ -19,103 +19,187 @@
...
+            track.stop();
+        });
     }
+
+    // remove newly created elements
+    removeAppendedCameraElements(video, customElements);
+}
+
+function removeAppendedCameraElements (video, customElements) {
+
```
erisu:
```suggestion
```

baibhavprdn:
Any idea when this feature will be included? This is exactly the thing I've been needing in my app.

README.md
breautek: `customCameraContainer` sounds like is something you need a specific `DOMElement`, so allowing flexibility of `querySelector` I think promotes bad coding practice.

If you need a specific element, the element should be tagged by an ID as to guarantee that you will receive the expected `DOMElement`. Using `querySelector` makes code brittle by picking the first occurrence of the selected term, which may not be the node you want.

The YouTube JS API[1] for example does something similar where it needs to insert HTML in a specific node of your choice. Their API signature accepts either an id, or the `HTMLElement` itself.

So I would prefer to keep `customCameraContainer` accepting an `id` rather than a query string. For flexibility, it could also optionally accept a DOM node and it would be up to the developer to find that reference as they see fit (which they could use `querySelector` if they wanted to, or it could be a fresh node that isn't even in the DOM tree yet (via `document.createElement`), etc.

[1] https://developers.google.com/youtube/iframe_api_reference#Loading_a_Video_Player

erisu: I believe this PR is actually no longer valid.

There were some discussion after the PR creation and a decision that users should maybe use WebRTC camera functionality over this plugin. After that, this PR development was stopped. Maybe this PR can be closed.

If memory serves, iOS has some limitation/problems with WebRTC camera functionality so the plugin will still be needed. It can be conditionally wrapped to decided if it use WebRTC over Plugin.

breautek: Must be this limitation:

> Does not work in standalone running ("installed") PWAs, getUserMedia returns no video input devices in UIWebView or WKWebView, but only directly in Safari.

https://caniuse.com/#search=getUserMedia

timbru31:
@baibhavprdn - No. As you can see, there are still unresolved discussions that need to be worked on first. Generally: No ETA on features, bugs and new releases.

Issue #492, opened 2019-08-23
iOS 13 camera is not working as expected


vlinde:
# Bug Report

## Problem
On iOS 13 the camera is not working as expected.

### What is expected to happen?
To save the image with content in the temp folder and to return the location.

### What does actually happen?
The image is saved and the plugin returns the path but the image content is missing.

### Command or Code
I use the following settings to open the camera.
```
{
      sourceType: camera.PictureSourceType.CAMERA,
      destinationType: camera.DestinationType.FILE_URI,
      quality: 92,
      encodingType: camera.EncodingType.JPEG,
      correctOrientation: true,
      saveToPhotoAlbum: false
}
```
When I use `camera.DestinationType.DATA_URL` instead of `camera.DestinationType.FILE_URI` I get the base64 code.

### Environment, Platform, Device
Tested on iOS 13 beta 6 & 7.

### Version information
```
Ionic:

   ionic (Ionic CLI)  : 4.12.0 (/usr/local/lib/node_modules/ionic)
   Ionic Framework    : ionic-angular 3.9.5
   @ionic/app-scripts : 3.2.4

Cordova:

   cordova (Cordova CLI) : 8.1.2 (cordova-lib@8.1.1)
   Cordova Platforms     : android 6.4.0, ios 5.0.1
   Cordova Plugins       : cordova-plugin-ionic-keyboard 2.1.3, cordova-plugin-ionic-webview 4.1.1, (and 34 other plugins)

System:

   ios-deploy : 1.9.4
   ios-sim    : 8.0.1
   NodeJS     : v11.11.0 (/usr/local/Cellar/node/11.11.0/bin/node)
   npm        : 6.7.0
   OS         : macOS Mojave
   Xcode      : Xcode 11.0 Build version 11M392q

```

## Checklist

- [X] I searched for existing GitHub issues
- [X] I updated all Cordova tooling to most recent version
- [X] I included all the necessary information above

vlinde:
I think there is a bug in iOS 13.
On iOS 13.1 and xCode 11.0 beta 7, the issue is still available.
Here is the fix: https://github.com/vlinde/cordova-plugin-camera-with-exif/commit/4afa1516b0037768a4c51dfc4c77a9d04f68b886
Is another plugin but the fix is the same.

mevinDhun:
@vlinde I made a PR to fix this:
https://github.com/apache/cordova-plugin-camera/pull/501

vlinde:
[...]

PR #501, opened 2019-09-13
fix(ios): corrupted image on ios 13

l0stpenguin:
### Platforms affected
ios 13.0, 13.1

### Motivation and Context

When taking an image on ios 13 via camera, the path is returned correctly but the temp image is corrupted.
Here is the issue:
https://github.com/apache/cordova-plugin-camera/issues/492

### Description

Upon debugging, i found that the image data in `self.data` in `CDVCamera.m` is being released/corrupted after setting the metadata via [`imagePickerControllerReturnImageResult`](https://github.com/apache/cordova-plugin-camera/blob/master/src/ios/CDVCamera.m#L678)
The solution is to create a copy of the `NSData` and use that to write the image on disk.

### Testing

Open camera on ios 13. Take an image. The image returned should be displayed correctly

### Checklist


src/ios/CDVCamera.m
```diff
@@ -670,17 +670,21 @@ - (void)imagePickerControllerReturnImageResult
 {
     CDVPictureOptions* options = self.pickerController.pictureOptions;
     CDVPluginResult* result = nil;
+   
+    NSMutableData *dest_data = [NSMutableData data];
```
timbru31: please use camelCase for variable names. Besides that, can we find a better name than `dest_data`?

mevinDhun: @timbru31 i renamed it to `imageDataWithExif`. is that fine for you?

mevinDhun:
> Thanks a lot for the PR and the debugging! Some minor comments from my side.
> 
> Besides that, does the change work on lower iOS versions, too?

Yes i have tested it on ios 12 too.

amayvs:
@mevinDhun @timbru31 Is there any ETA on merging this PR?

anubhavberry:
I've taken the latest from the fork and tried to test on 13.1.3. The photo doesn't show up

eastwillie:
I've tried on 13.1.3 wkWebview, it didn't work.

mevinDhun:
@eastwillie @anubhavberry i have tested it on ios 13.1 and 13.2 and confirm it is working properly.

eastwillie:
@mevinDhun 
Here is my code and after I take a photo, there was an error:
finalize:2667: image destination must have at least one image.
Seems to me that the picture hasn't been saved into tmp somehow? Or I made a config mistake?

`openCamera(camera = true) {
            const options = {
                quality: 50,
                targetWidth: 750,
                targetHeight: 750,
                destinationType: window.Camera.DestinationType.FILE_URI,
                sourceType: camera ? window.Camera.DestinationType.FILE_URI : window.Camera.PictureSourceType.SAVEDPHOTOALBUM,
                encodingType: window.Camera.EncodingType.JPEG,
                mediaType: window.Camera.MediaType.PICTURE,
                allowEdit: true,
                correctOrientation: true,
            };
            navigator.camera.getPicture((imageUri) => {
                this.$emit('upload', {
                    success: true,
                    uri: [imageUri],
                });
            }, () => {
                this.$emit('upload', {
                    success: false,
                    uri: [],
                });
            }, options);
        },`

ryaa:
I have been experience the same problem on iOS 13.2 and the problem seems to be that when we copy the image with metadata (exif data) from the source to destination in imagePickerControllerReturnImageResult method of CDVCamera.m the resulting file is of 0 size and CGImageDestinationFinalize returns false (fails!). If I disable geolocation for camera by setting the below in config.xml and everything works well.
```
<preference name="CameraUsesGeolocation" value="false" />
```

I tried two fixes and both seems to be working:
1) I changed imagePickerControllerReturnImageResult in CDVCamera.m based on this suggestion https://github.com/vlinde/cordova-plugin-camera-with-exif/commit/99465b4c078bac48f26eb28a90737396e7dfca92

I removed
```
NSData* data_content = [self.data mutableCopy];
```
and added instead
```
UIImage *image = [UIImage imageWithData:self.data];
CGImageRef imageRef = image.CGImage;
```
plus I replaced this line
```
 CGImageDestinationAddImageFromSource(destinationImage, sourceImage, 0, (__bridge CFDictionaryRef)self.metadata);
```
with this line
```
CGImageDestinationAddImage(destinationImage , imageRef, (CFDictionaryRef)self.metadata);
```
2) I also tried the fix as suggested by @mevinDhun 

Both fixes seems to be working.

I'm still wondering why CGImageDestinationAddImageFromSource fails while CGImageDestinationAddImage works fine when saving the image with the metadata. Any ideas?

eastwillie:
@ryaa Thanks, I saw a dialog ask permission for location, but less than 1 second and disappeared. That explained it.

erisu (changes requested):
The changes in `package.json` and `plugin.xml` should be reverted. Also the branch will need to be rebased with master as there is conflicting files.

erisu:
Closed by #712

Issue #416, opened 2019-01-28
Cannot take pictures on Android


Owlbertz:
Hi,

I am using the camera plugin to take pictures and to select photos from the gallery. While the latter works perfectly fine, the former action fails with this error:

`length=1; index=1`

I was able to find out the source, but not the cause:

```
"F09 Camera1798421768 sjava.lang.ArrayIndexOutOfBoundsException: length=1; index=1
	 at org.apache.cordova.camera.CordovaUri.getFileNameFromUri(CordovaUri.java:98)
	 at org.apache.cordova.camera.CordovaUri.<init>(CordovaUri.java:52)
	 at org.apache.cordova.camera.CameraLauncher.takePicture(CameraLauncher.java:330)
	 at org.apache.cordova.camera.CameraLauncher.callTakePicture(CameraLauncher.java:308)
	 at org.apache.cordova.camera.CameraLauncher.execute(CameraLauncher.java:184)
	 at org.apache.cordova.CordovaPlugin.execute(CordovaPlugin.java:98)
	 at org.apache.cordova.PluginManager.exec(PluginManager.java:132)
	 at org.apache.cordova.CordovaBridge.jsExec(CordovaBridge.java:59)
	 at org.apache.cordova.engine.SystemExposedJsApi.exec(SystemExposedJsApi.java:41)
	 at android.os.MessageQueue.nativePollOnce(Native Method)
	 at android.os.MessageQueue.next(MessageQueue.java:323)
	 at android.os.Looper.loop(Looper.java:136)
	 at android.os.HandlerThread.run(HandlerThread.java:61)
```

Arguments provided were to the Java-file were:
0:	70
1:	1
2:	1
3:	1024
4:	-1
5:	0
6:	0
7:	false
8:	true
9:	false
10:	null
11:	0

I hope someone has an idea what I am doing wrong.

hazems:
How do you launch camera, can you share your code? What is your Android device?

pvillaverde:
Hey there!

I'm having the same issue using Ionic. It works when using library but if I try to use the camera it gives the  same error. 
Here's the code I used:

```javascript

function takePicture(camera) {
	const options = {
		//	quality: 90, //Number	Quality of the saved image, range of 0 - 100
		destinationType: $cordovaCamera.DestinationType.FILE_URI, //Number	Format of the return value
[...]

PR #510, opened 2019-10-11
(android) fix: use provider prefix to avoid conflict with cordova-open plugin

LoicEvenium:
### Platforms affected
cordova-android

### Motivation and Context
Application crash when open the camera if `cordova-plugin-open` installed.
Providers of `cordova-plugin-open` and `cordova-plugin-camera` are in conflicts using the same `android:authorities` value. See the AndroidManifest:
```xml
    <provider android:authorities="${applicationId}.provider" android:exported="false" android:grantUriPermissions="true" android:name="org.apache.cordova.camera.FileProvider">
      <meta-data android:name="android.support.FILE_PROVIDER_PATHS" android:resource="@xml/camera_provider_paths"/>
    </provider>
    <provider android:authorities="${applicationId}.provider" android:exported="false" android:grantUriPermissions="true" android:name="androidx.core.content.FileProvider">
      <meta-data android:name="android.support.FILE_PROVIDER_PATHS" android:resource="@xml/provider_paths"/>
    </provider>
```
Fix: #416

### Description
Add "camera" in the `android:authorities` to void namespace conflict

### Testing
Open camera, take a picture

### Checklist


pseudooriginal:
Can this PR be merged please? It brakes the plugin on all Android devices.

plugin.xml
```diff
@@ -59,7 +59,7 @@
         <config-file target="AndroidManifest.xml" parent="application">
           <provider
               android:name="org.apache.cordova.camera.FileProvider"
-              android:authorities="${applicationId}.provider"
+              android:authorities="${applicationId}.camera.provider"
```
breautek, line 62: If these can conflict with other plugins I think I'd prefer to stay away from generic terms such as `camera`, otherwise I feel like there is still a decent risk of running into this kind of problem in the future.

Therefore I think something like `"${applicationId}.cordova.plugin.camera.provider"` will be better, as it provides a pretty specific context.

```suggestion
              android:authorities="${applicationId}.cordova.plugin.camera.provider"
```

src/android/CameraLauncher.java
```diff
@@ -301,7 +301,7 @@ public void takePicture(int returnType, int encodingType)
         // Specify file so that large image is captured and returned
         File photo = createCaptureFile(encodingType);
         this.imageUri = new CordovaUri(FileProvider.getUriForFile(cordova.getActivity(),
-                applicationId + ".provider",
+                applicationId + ".camera.provider",
```
breautek, line 304: Same as above

```suggestion
                applicationId + ".cordova.plugin.camera.provider",
```

src/android/CameraLauncher.java
```diff
@@ -799,7 +799,7 @@ else if (srcType == CAMERA) {
                 try {
                     if (this.allowEdit) {
                         Uri tmpFile = FileProvider.getUriForFile(cordova.getActivity(),
-                                applicationId + ".provider",
+                                applicationId + ".camera.provider",
```
breautek, line 802: same as above

```suggestion
                                applicationId + ".cordova.plugin.camera.provider",
```

pseudooriginal:
Thank you for your speed 😊

plugin.xml
```diff
@@ -59,7 +59,7 @@
         <config-file target="AndroidManifest.xml" parent="application">
           <provider
               android:name="org.apache.cordova.camera.FileProvider"
-              android:authorities="${applicationId}.provider"
+              android:authorities="${applicationId}.cordova.plugin.camera.provider"
```
timbru31, line 62: Is this considered a breaking change for existing applications?

breautek: Maybe....

https://developer.android.com/guide/topics/manifest/provider-element

Looks like this provider authority name is used in `content://` urls, but I believe these are just temporarily paths anyway...? Wouldn't you read the path and save it to a more persistent location, generally speaking?

netzzeit:
Is there any chance this change is being merged soon? :)

timbru31:
There is still an open discussion. I'm not merging something that I'm not 100% sure of that it does not cause a breaking change - unless it's explicitly stated so.

plugin.xml
jcesarmobile: I don't think it's breaking, but we are doing a major release anyway, so it won't be a problem anyway. I fixed some conflicts it had, going to merge.

We should probably add a provider in cordova-android for plugins to use instead of each plugin having it's own, but providers need an Android support/Android X dependency, so not sure if we want that in cordova-android.
If not, just a simple official plugin that only have the provider and others can use as a dependency.

Issue #545, opened 2020-01-13
[iOS - iPad] App crashes if video transcoding is interrupted


guillem-sopra:
# Bug Report

## Problem

In iOS, when one selects a video from the gallery, a transcoding/compression process must be carried out. In iPhone one cannot interrupt or cancel this process, but in iPad you do: since the gallery navigation appears inside a pop-up, you can click anywhere outside the pop-up or the back button. If you do so, the app crashes.

### What is expected to happen?

The transcoding process is interrupted, an exception is captured and properly emitted in js.

### What does actually happen?

The app crashes. My guessing: An uncaptured exception happens in the native code.

## Information

Tested in iOS 12 and 13.

### Command or Code

The app crashes and the catch is never reached:

```
fileUri = await this.camera.getPicture(options).then((fileURI) => {
      console.log(fileURI);
      return fileURI;
    }).catch((error) => {
    console.error(error)
  });
```

### Environment, Platform, Device

iPad, iOS versions 12 and 13.

## Checklist

- [x] I searched for existing GitHub issues
- [x I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

PR #547, opened 2020-01-16
fix: [iOS - iPad] App crashes if video transcoding is interrupted

guillem-sopra:
Closes https://github.com/apache/cordova-plugin-camera/issues/545


guillem-sopra:
OK, I'll close this one

PR #588, opened 2020-04-28
[Android] Cache images in device storage, devices have enough space now.

purplecabbage:
### Platforms affected
Android

### Motivation and Context

We used to have to cache images on removable storage because devices did not have enough space, this is not an issue any longer, and it is better to let the system handle cache cleanup.

### Description

Remove conditional, and just default to application storage

### Checklist


JacobIsrael18:
Google's camera app and MANY others are still crashing or freezing.
Screenshot_20210117-000351_Camera

Issue #597, opened 2020-05-13
Android: Camera fails with an exception


ryaa:
# Bug Report

### What is expected to happen?
Camera should take photos as expected.

### What does actually happen?
An exception is thrown
```
java.lang.IllegalArgumentException: Failed to find configured root that contains /data/data/<PACKAGE_NAME>/cache/.Pic.jpg
```

## Information

The problem might be related to this commit https://github.com/apache/cordova-plugin-camera/commit/df1441420376696994d77f0856ab3dfbed5f0bbe

The plugin uses this configuration for FileProvider from **android/xml/camera_provider_paths.xml** with external-path however it might need to use cache-path instead.

I will try to provide a PR

### Command or Code

Install the latest version 4.2.0 which includes the above commit/change

### Environment, Platform, Device

Android 10

### Version information

N/A

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

ryaa:
The potential changes to fix the above and EXIF related problem in my PR https://github.com/apache/cordova-plugin-camera/issues/524

matiaspando:
Hey guys!  I have the same problem , is there any  workaround to avoid this issue ?

timbru31:
Use 4.1.0 and not an unreleased, unstable version.

ryaa:
> Hey guys! I have the same problem , is there any workaround to avoid this issue ?

Until the fix is merged/provided, you can install from my fork https://github.com/ryaa/cordova-plugin-camera where this and some other problems have been fixed.

CodeWithOz:
@ryaa thanks for your fork, it fixes the issue on my end. However, have you noticed an increase in the size of your app bundle as a result of using your fork? My bundle went from 16.8MB to 23MB after switching to your fork. Any idea why that is?

CodeWithOz:
I'm also seeing a bug where my app gets shifted by ~50px at the bottom. The black area did not exist previously.
![blank-space](https://user-images.githubusercontent.com/28525986/86926214-b6a04c80-c129-11ea-92a4-9d627c1f9880.png)

I noticed that I use CSS to reposition it correctly, but this wasn't happening before I installed your fork. Any ideas?

PR #617, opened 2020-06-26
breaking(android): stop using CordovaUri helper class

jcesarmobile:
### Platforms affected
android

### Motivation and Context
At the moment the plugin tries to guess (in a dirty way) the image path from the content url, but that's not really necessary since we know the path beforehand, and doesn't work at the moment since the path was changed in a previous PR causing #597 

### Description

probably breaking change since CordovaUri class says it's for Android 4.4.4 and lower support, but since cordova-android 9 won't support versions older than 5.1, I think we can remove the class and we also need to update the plugin to use Android X, which will require cordova-android 9.

What the PR does is to store the file path in a variable and use that variable to send the camera result instead of trying to guess the image path from the returned result from the camera intent, which was a dirty way of getting it (according to the method description) and which doesn't work since [this](https://github.com/apache/cordova-plugin-camera/commit/df1441420376696994d77f0856ab3dfbed5f0bbe) was merged.

For that, it removes CordovaUri.java class and it's usages.
Saves image paths in a variable to return.
Changes the file_provider to use `cache_files` as the mentioned commit stores the images there.
Removes code used for Android 4.4 and older.

fixes https://github.com/apache/cordova-plugin-camera/issues/597

### Testing
Tested on Android 5 and Android 9 devices with this code

```
navigator.camera.getPicture(onSuccess, onFail, { quality: 50, destinationType: Camera.DestinationType.FILE_URI });
    function onSuccess(imageURI) {
        alert(imageURI);
        document.getElementById("photoResult").src = imageURI;
    }
    function onFail(message) {
        alert('Failed because: ' + message);
    }
```

### Checklist


src/android/CameraLauncher.java
```diff
@@ -290,10 +292,13 @@ public void takePicture(int returnType, int encodingType)
 
         // Specify file so that large image is captured and returned
         File photo = createCaptureFile(encodingType);
-        this.imageUri = new CordovaUri(FileProvider.getUriForFile(cordova.getActivity(),
+        System.out.println("photo "+photo.getAbsolutePath());
```
erisu, line 295: Is this needed?

src/android/CameraLauncher.java
```diff
@@ -290,10 +292,13 @@ public void takePicture(int returnType, int encodingType)
...
+        this.imageFilePath = photo.getAbsolutePath();
+        this.imageUri = FileProvider.getUriForFile(cordova.getActivity(),
                 applicationId + ".provider",
-                photo));
-        intent.putExtra(MediaStore.EXTRA_OUTPUT, imageUri.getCorrectUri());
+                photo);
+        System.out.println("image uri "+this.imageUri);
```
erisu, line 300: Same

erisu:
Other then the comments, I think the rest is OK 👌

erisu:
You might also be able to rebase with master to get the additional tests to pass.

PR #627, opened 2020-07-14
breaking: bump version 5.0.0-dev

erisu:
### Motivation and Context

Bump for Major Release

PR #628, opened 2020-07-14
breaking: bump project requirements

erisu:
### Motivation and Context

Bump the project requirements.

- Cordova Android >= 9.0.0
- Cordova CLI >= 9.0.0

**Note:** Changes also include #627. This PR should be rebased after #627 is merged in.


erisu:
@jcesarmobile Any thoughts on iOS requirements being set to 6.x? 

I don't know if there are any breaking changes on the iOS side of this plugin, but from the current testing system standpoint (paramedic) I believe it always checks out the latest Cordova iOS (6.x) which requires Xcode 11 to be used. This means we might be only testing Xcode 11+... At least the additional tests do not pass on any version below Xcode 11.

jcesarmobile:
I think we should require cordova-ios 5.1.0, I see no reason to require 6.0.0 in this plugin.

Not being able to test on Xcode 10 is not a big problem, considering Apple requires Xcode 11 to submit apps.

I've created a thread on the mail list to talk about the engines and used this PR as an example of what I think we should be doing.

erisu:
@jcesarmobile I set it to `cordova-ios >= 5.1.0` as you mentioned in previous comment but also debated if it should really be `>= 5.1.1`. `5.1.0` had a bug with the new WKWebViewOnly flag being and prepare ran before the platform existed. Or something like this if memory is correct. But `5.1.1` fix that prepare issue surrounding the flag.

Other then that, this PR I think is OK.

jcesarmobile:
I guess if people face the prepare bug will update from 5.1.0 to 5.1.1, but you can set it to 5.1.1 if you prefer

PR #634, opened 2020-07-23
chore: remove deprecated file transfer plugin

erisu:
### Platforms affected

iOS

### Motivation, Context & Description

Remove the broken deprecated File Transfer Plugin from test & its test,


jcesarmobile:
I loaded the url and says Hello!, so not sure if broken, didn't try to upload a picture, but it's not down

timbru31:
Oh indeed - my PiHole blocked the whole Heroku domain. Yeah it seems it's still alive. Approval remains though.

PR #637, opened 2020-07-24
breaking: remove NATIVE_URI DestinationType

jcesarmobile:
### Platforms affected
all

### Motivation and Context

`NATIVE_URI` was only used on iOS and windows, on Android it acts like `FILE_URI`.

On iOS it returned `ALAssetsLibrary` urls that are not compatible with `WKWebView` and `ALAssetsLibrary` was deprecated on iOS 9, so I don't think it makes sense to keep this destination type since it's not supported by all platforms and it's deprecated on iOS.

### Description

also removes other usages of `ALAssetsLibrary` since it's deprecated

### Testing

tested on real iOS device

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,19 +276,14 @@ Optional parameters to customize the camera settings.
 
 ### Camera.DestinationType : <code>enum</code>
 Defines the output format of `Camera.getPicture` call.
-_Note:_ On iOS passing `DestinationType.NATIVE_URI` along with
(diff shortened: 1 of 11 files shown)


FBNitro:
This is used on Windows also I thought, because if I pick a file from the photo storage, I can only seem to get a hold of the correct path using the NATIVE_URI.  

I'm currently on 4.1.0.

Gillardo:
when this is removed, how do i get the PATH to the file from the PHOTOLIBRARY, as the FILE_URI returns the file that is in temporary storage, i want the file that is saved like NATIVE_URI did before

PR #641, opened 2020-08-04
Update empty filename issue for supporting multiple images

arshad54:
### Motivation and Context

I was stucked from few days because i am not able to get correct path for the images selected using PHOTOALBUM but then i got a new problem after rectifying the issue i found that the cameraLauncher.java file renaming every image selected from gallery to .Pic. therefore its very difficult to differentiate them and also it is overriding the images.

### Description

This is required to differ between multiple images as the java file renaming every file to .Pic. so instead this i added timestamp to the empty filename so that the developer can able to differentiate between multiple files and their paths.

### Testing

I have tested this on android 10 using camera and gallery app in my device and it is working fine.

### Checklist

- [x ] I've run the tests to see all new and existing tests pass

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -335,7 +335,9 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
      */
     private File createCaptureFile(int encodingType, String fileName) {
         if (fileName.isEmpty()) {
-            fileName = ".Pic";
+            //creating timestamps for empty filename to differentiate between multiple files selected via gallery
+            String timeStamp = new SimpleDateFormat(TIME_FORMAT).format(new Date());
+            fileName = timeStamp;
         }
 
         if (encodingType == JPEG) {


src/android/CameraLauncher.java
```diff
@@ -335,7 +335,9 @@ private File createCaptureFile(int encodingType) {
      */
     private File createCaptureFile(int encodingType, String fileName) {
         if (fileName.isEmpty()) {
-            fileName = ".Pic";
+            //creating timestamps for empty filename to differentiate between multiple files selected via gallery
+            String timeStamp = new SimpleDateFormat(TIME_FORMAT).format(new Date());
+            fileName = timeStamp;
```
erisu, line 340:
```suggestion
            // Append the current time in milliseconds for empty filename to differentiate between multiple files selected via gallery
            fileName = ".Pic-" + System.currentTimeMillis();
```

IMO, I would suggest the above.

What you have might work to a certain point.

The `TIME_FORMAT`'s lowest placement is in seconds. 

Could someone take burst shots and therefore have many photos per second. Using `System.currentTimeMillis` might help out a bit since it's now in milliseconds.

Example filename with the above code would be: `.Pic-1647402802991`

I would not recommend updating the value of `TIME_FORMAT`. If changed, then the PR becomes a major breaking PR.

To keep or not to keep the `".Pic-"` prefix is questionable? If it is kept as a prefix, then it could be used as a keyword filter when grabbing all files with the `.Pic` prefixed. I don't know if this is necessary, but I also don't know what other files exist in the same directory.

What do you think?

PR #654, opened 2020-09-03
chore: add release notify action

NiklasMerz:
### Motivation and Context
We want to test a new action that closes issues that are just asking for a new release and reduce noise. Let's start with this plugin.

### Description

This PR adds a new Github action built by me to add a comment and closes the issue if it looks like the issue is just created to ask for a release.

### Checklist

### DIFF
--- a/.github/workflows/release-notify.yml
+++ b/.github/workflows/release-notify.yml
@@ -3,7 +3,6 @@ name: Close issue asking for release
 on:
   issues:
     types: [opened]
-    
 
 jobs:
   action-test:
--- a/.github/workflows/release-notify.yml
+++ b/.github/workflows/release-notify.yml
@@ -3,8 +3,6 @@ name: Close issue asking for release
 on:
   issues:
     types: [opened]
-  issue_comment:
-    types: [created]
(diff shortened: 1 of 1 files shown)


NiklasMerz:
Some discussion here: https://github.com/NiklasMerz/release-notify/issues/4

.github/workflows/release-notify.yml
```diff
@@ -0,0 +1,16 @@
...
+    types: [created]
+    
+
+jobs:
+  action-test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: niklasmerz/release-notify@master
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
```
timbru31, line 16: Won't we need infra to add the secret?

NiklasMerz: https://docs.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token

I did not find an option to disable it in my orgs. Not sure though.

erisu:
👍

Issue #471, opened 2019-06-27
[Android] Permission denied when trying to launch camera


ilyakamens:
# Bug Report

## Problem

I'm unable to launch the camera on Android.

### What is expected to happen?

The camera should launch.

### What does actually happen?

The error callback is called with a value of `20`, which I believe is `PERMISSION_DENIED_ERROR`: https://github.com/apache/cordova-plugin-camera/blob/master/src/android/CameraLauncher.java#L98

## Information

It's the only call I make on the page (after device is ready, of course).

### Command or Code

```
window.navigator.camera.getPicture(
  (data: any) =>  alert(`success: ${data}`),
  (err: any) => alert(`error: ${err}`),
);
```

### Environment, Platform, Device

I've tried on an LG K20 PLUS running Android 7.0 and a Galaxy S9 running Android 9.

### Version information

```
$ cordova --version
9.0.0 (cordova-lib@9.0.1)
```

```
$ cordova platforms
Installed platforms:
  android 8.0.0
  ios 5.0.0
```

```
$ cordova plugins list
cordova-custom-config 5.1.0 "cordova-custom-config"
cordova-plugin-add-swift-support 2.0.1 "AddSwiftSupport"
cordova-plugin-advanced-http 2.0.6 "Advanced HTTP plugin"
cordova-plugin-android-referrer 1.0.0 "cordova-plugin-android-referrer"
cordova-plugin-androidx 1.0.2 "cordova-plugin-androidx"
cordova-plugin-androidx-adapter 1.0.2 "cordova-plugin-androidx-adapter"
cordova-plugin-appsflyer-sdk 4.4.15 "AppsFlyer"
cordova-plugin-backbutton 0.3.0 "Backbutton"
cordova-plugin-camera 4.0.3 "Camera"
cordova-plugin-device 2.0.2 "Device"
cordova-plugin-dialogs 2.0.1 "Notification"
cordova-plugin-facebook4 4.2.1 "Facebook Connect"
cordova-plugin-file 6.0.1 "File"
cordova-plugin-firebase 2.0.5 "Google Firebase Plugin"
cordova-plugin-geolocation 4.0.1 "Geolocation"
cordova-plugin-globalization 1.11.0 "Globalization"
cordova-plugin-inappbrowser 3.1.0-dev "InAppBrowser"
cordova-plugin-ionic-webview 4.0.1 "cordova-plugin-ionic-webview"
cordova-plugin-keyboard 1.2.0 "Keyboard"
[...]

PR #670, opened 2020-10-01
fix(android): Declare CAMERA permission

breautek:
### Platforms affected
Android

### Motivation and Context

Fixes https://github.com/apache/cordova-plugin-camera/issues/471

Fixes permission issue on Android.

Camera permission is generally only required if you're using the Camera APIs, which we don't however there are a few exceptions to this rule.

As of Android M, if you try to request the `CAMERA` permission without declaring it in the manifest, a security error is thrown. See the [Note](https://developer.android.com/reference/android/provider/MediaStore.html#ACTION_IMAGE_CAPTURE) in the android documentation.

The camera plugin does request the `CAMERA` permission to handle an edge case where another plugin may include it in the manifest.

https://github.com/apache/cordova-plugin-camera/blob/5ae56cf8f07bcdfa090d928e7af3c8637521afce/src/android/CameraLauncher.java#L249-L251

### Description

Adds `<uses-permission android:name="android.permission.CAMERA" />` to `plugin.xml`

### Testing

- Ran `npm test`.
- Tested on one of my production apps.

### Checklist

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -56,6 +56,7 @@
         </config-file>
         <config-file target="AndroidManifest.xml" parent="/*">
             <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+            <uses-permission android:name="android.permission.CAMERA" />
         </config-file>
         <config-file target="AndroidManifest.xml" parent="application">
           <provider


erisu (approved):
I am OK with adding the permission.

NiklasMerz (approved):

breautek:
The failed tes in travis is caused by unrelated platforms, so merging.

Thanks for the reviews.

jcesarmobile:
The plugin is supposed to request the permission only if the permission is present, if that's not working properly it should be fixed instead of adding the permission. 
If we are adding the permission then the code that checks if the permissions is present should be removed because it will always be present, but I think we should revert this, I don't think it's not good to add unneeded permissions.

breautek:
I can revisit this and try to determine why it was required for the camera permission to be declared in my app. My original conclusion was based on my own experience and several other reports that declaring the permission was necessary.

jcesarmobile:
In the issue linked people talk about <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />, they removed it somehow and had to add it, it’s supposed to be added by the plugin but some had hooks that removed it.
One user ask if camera permission is missing and another one says adding camera permission didn’t fix their problems.

breautek:
> In the issue linked people talk about , they removed it somehow and had to add it, it’s supposed to be added by the plugin but some had hooks that removed it.
> One user ask if camera permission is missing and another one says adding camera permission didn’t fix their problems.

Yes, and it looks like in my case it's my app to blame here. I had a deeper look and it looks like the app is using a permissions plugin to explicitly request the camera permission (then failing because the permission is not declared in the manifest). In reality, my app shouldn't even need to request camera permissions.

I'll prepare a revert PR.

PR #677, opened 2020-10-14
Revert "fix(android): Declare CAMERA permission"

breautek:
### Platforms affected
Android

### Motivation and Context

Reverts https://github.com/apache/cordova-plugin-camera/pull/670

The blame is actually not with the plugin, but rather the app. Camera plugin will request camera permissions if the camera permission is declared to support coexisting with other plugins that may need camera permission APIs, but otherwise the plugin does not require the camera permission to be declared and therefore it should not declare itself.

If the app uses a plugin that requires the camera permission, it is up to that plugin to add the camera permission itself.

### Description

This reverts commit 140e8861e30164da0861fec548d0aa2413197e12.

### Testing

Ran `npm test`

### Checklist

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -56,7 +56,6 @@
         </config-file>
         <config-file target="AndroidManifest.xml" parent="/*">
             <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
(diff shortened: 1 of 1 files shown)


jcesarmobile (approved):
LGTM

PR #687, opened 2020-11-09
fix(android): return exception message (where it exists) 

dpa99c:
### Platforms affected
Android

### Motivation and Context

I use this plugin in multiple apps and sometimes users encounter errors when using the camera and ask for support.
To help with user support it would be useful to know the cause of the error by returning the exception message (where it exists) in the error message sent to the plugin error callback.
For example, this can be logged to analytics and used to help diagnose the cause of the error.

### Description

Where an exception is raised and the plugin invokes the error callback, append the exception message to the returned error message.

### Testing

Tested on Android device by forcing an exception by appending "FOO" to the `imageUri`, i.e. change [line 615](https://github.com/apache/cordova-plugin-camera/blob/master/src/android/CameraLauncher.java#L615) to:

	InputStream fileStream = org.apache.cordova.camera.FileHelper.getInputStreamFromUriString(imageUri.toString()+"FOO", cordova);

Resulting error callback message: `Error capturing image: open failed: ENOENT (No such file or directory)`	

### Checklist

- [n/a] I added automated test coverage as appropriate for this change
- [n/a] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [n/a] I've updated the documentation if necessary

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -759,7 +759,7 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
 
                             } catch (Exception e) {
                                 e.printStackTrace();
-                                this.failPicture("Error retrieving image: "+e.getMessage());
+                                this.failPicture("Error retrieving image: "+e.getLocalizedMessage());
                             }
                         } else {
                             this.callbackContext.success(finalLocation);
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -829,7 +829,7 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
                     }
                 } catch (IOException e) {
                     e.printStackTrace();
-                    this.failPicture("Error capturing image: "+e.getMessage());
+                    this.failPicture("Error capturing image: "+e.getLocalizedMessage());
                 }
             }
 
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -1268,7 +1268,7 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
                 code = null;
             }
         } catch (Exception e) {
-            this.failPicture("Error compressing image: "+e.getMessage());
+            this.failPicture("Error compressing image: "+e.getLocalizedMessage());
         }
         jpeg_data = null;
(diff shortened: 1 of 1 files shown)


erisu (approved):
LGTM 👍

src/android/CameraLauncher.java
```diff
@@ -759,7 +759,7 @@ else if (destType == FILE_URI) {
 
                             } catch (Exception e) {
                                 e.printStackTrace();
-                                this.failPicture("Error retrieving image.");
+                                this.failPicture("Error retrieving image: "+e.getMessage());
```
jcesarmobile, line 762:
```suggestion
                                this.failPicture("Error retrieving image: "+e.getLocalizedMessage());
```

src/android/CameraLauncher.java
```diff
@@ -829,7 +829,7 @@ else if (srcType == CAMERA) {
                     }
                 } catch (IOException e) {
                     e.printStackTrace();
-                    this.failPicture("Error capturing image.");
+                    this.failPicture("Error capturing image: "+e.getMessage());
```
jcesarmobile, line 832:
```suggestion
                    this.failPicture("Error capturing image: "+e.getLocalizedMessage());
```

src/android/CameraLauncher.java
```diff
@@ -1268,7 +1268,7 @@ public void processPicture(Bitmap bitmap, int encodingType) {
                 code = null;
             }
         } catch (Exception e) {
-            this.failPicture("Error compressing image.");
+            this.failPicture("Error compressing image: "+e.getMessage());
```
jcesarmobile, line 1271:
```suggestion
            this.failPicture("Error compressing image: "+e.getLocalizedMessage());
```

jcesarmobile (changes requested):
In other parts of the plugin we are using `getLocalizedMessage` instead of `getMessage`, so for consistency I think we should also use `getLocalizedMessage` here.
I get the exact message since it's not localized.

https://github.com/apache/cordova-plugin-camera/blob/ebe0517a243b7a710d225569ce7a228c346c77f7/src/android/CameraLauncher.java#L944

PR #688, opened 2020-11-12
Platform Android support for AndroidX

sieman:
<!--
Please make sure the checklist boxes are all checked before submitting the PR. The checklist is intended as a quick reference, for complete details please see our Contributor Guidelines:

http://cordova.apache.org/contribute/contribute_guidelines.html

Thanks!

### Platforms affected
Only Android

### Motivation and Context

This Change is required to support new Version of Andoid. AndroidX put old support libraries into a new namespace.

### Description

It is a renaming of dependencies. Wherever the support library was imported, I replace it with the new namespace. 

### Testing
I tested the camera plugin within the app of my employer and it works.


timbru31:
Thanks Siegmar for your PR. We've already discussed AndroidX a bit and it's a tricky decision. On the one hand, if we go for AndroidX and enforce it, we will probably break a lot of plugins. But on the other hand, there might be some changes by Google that forces us to enable AndroidX nevertheless.

In the meantime you could use the cordova-plugin-androidx-adapter (which you probably already do).

See also https://github.com/apache/cordova-android/issues/841 and the discussion on YT: https://www.youtube.com/watch?v=oarY6ZYj-CU, if interested.

noahcooper:
Any update on when this plugin might move to AndroidX? I'm sure I'm not alone in saying this -- this is now the only remaining plugin my apps use that requires the use of `cordova-plugin-androidx-adapter`.

plugin.xml
```diff
@@ -79,9 +79,10 @@
         <js-module src="www/CameraPopoverHandle.js" name="CameraPopoverHandle">
             <clobbers target="CameraPopoverHandle" />
           </js-module>
-
-        <preference name="ANDROID_SUPPORT_V4_VERSION" default="27.+"/>
-        <framework src="com.android.support:support-v4:$ANDROID_SUPPORT_V4_VERSION"/>
+        <preference name="AndroidXEnabled" value="true" />
```
NiklasMerz, line 82: As cordova-android 10 now comes with AndroidX by default those aren't really needed I think.

NiklasMerz (changes requested):

NiklasMerz:
As cordova-android 10 now works only with AndroidX we should move forward with this.

jcesarmobile:
It was done in https://github.com/apache/cordova-plugin-camera/pull/751, closing

PR #715, opened 2021-03-21
(iOS) overlay view supporting to show the indicated image titled "overl…

hajunho:
…ayImage.png"

### Platforms affected
Nothing.

### Motivation and Context

Another Cordova Plug-in uses a Cordova camera also. Overlay view is the primary API of the iOS platform. I think the overlay view function is inserted into the Cordova primary function to follow the trend of native Apps.

None.

### Testing

I have several projects used by over 1000+ users of the company titled Daelim. I've tested with this plugin and tested the hello-world program. As you can see from the codes, It is just for the iOS user to overlay view. I think it starts at the codes to guide. Will uses very usefully later on. I think I can update & test every month for this.

### Checklist

### DIFF
--- a/src/ios/CDVCamera.h
+++ b/src/ios/CDVCamera.h
@@ -89,6 +89,8 @@ typedef NSUInteger CDVMediaType;
 @property (strong) NSMutableDictionary *metadata;
 @property (strong, nonatomic) CLLocationManager *locationManager;
 @property (strong) NSData* data;
+@property (assign) BOOL isOverlayView;
+@property (strong) UIView* overlayView;
 
 /*
  * getPicture
@@ -101,6 +103,8 @@ typedef NSUInteger CDVMediaType;
  *	quality: integer between 1 and 100
  */
 - (void)takePicture:(CDVInvokedUrlCommand*)command;
+- (void)takePictureOverlayView:(CDVInvokedUrlCommand*)command;
+
 - (void)cleanup:(CDVInvokedUrlCommand*)command;
 - (void)repositionPopover:(CDVInvokedUrlCommand*)command;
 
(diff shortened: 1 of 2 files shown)


timbru31:
Thanks for your PR - could you please elaborate what it does? It's not quite clear to me what addition it has or bug it fixes. Besides that, it would be good to keep the different platform all in sync, i.e., adding an Android implementation, too. (if suitable & feasable)

hajunho:
> Thanks for your PR - could you please elaborate what it does? It's not quite clear to me what addition it has or bug it fixes. Besides that, it would be good to keep the different platform all in sync, i.e., adding an Android implementation, too. (if suitable & feasible)

Thank you for reviewing. I didn't touch the established codes. It's just an indirection layer. As you already know, the overlay view can be used for the additional view to putting the picture after the function "taking the picture". We use this for the watermark for that(I had adjusted Alpha value 0.38). Platform-specific codes are not applicable, I know. but I agree with your opinion to add this function to Android also. These codes have to be changed after using the additional function by the native developer. What I didn't decide things is what options have to be inserted. i.e Alpha value, images, location, size, watermark or location, or timestamp options. Those are very big considering but I or somebody have to do I think. These are base codes for that and no errors(No called basically)

IllusionVK:
Hey @hajunho! I have read all the comments on this, but i'm still not clear on what it does.  Do your changes allow for the use of this? https://developer.apple.com/documentation/uikit/uiimagepickercontroller/1619113-cameraoverlayview

And what exactly is that used for?  when I looked it seems it can allow custom buttons ? or maybe overlay on preview ? (see some examples below)

- Example 1: User is trying to show different buttons instead of normal iOS buttons:
https://stackoverflow.com/questions/12734407/uiimagepickercontroller-cameraoverlayview-not-working-with-ipho

- Example 2: user is trying to show a shape on the screen to help the user take a photo in the right location:
https://github.com/jamesmontemagno/MediaPlugin/issues/377

hajunho:
Hello,

I think this function is not needed to this point.
Please drop my codes. Actually I found that the overlayview project for
camera on github.

Thank you

PR #745, opened 2021-07-06
added github action workflow to run tests

purplecabbage:
### Checklist


erisu:
I rebased the PR and also removed the codecov step.

For codecov to work, we should have a testing framework that runs and generates code coverage reports. All of our plugins do not run any unit tests to create these reports. There is nothing to upload, so the best solution at this time is to remove the codecov step.

Issue #732, opened 2021-05-25
iOS : remove video compression on select from library


rudyZanotti:
Hello,

I would like to know if it is possible to remove compression when selecting video on iOS. My users use a lot of videos and I would like to compress my videos while they can do other things.

thank you in advance 
Rudy

PieterVanPoyer:
As far as I can see there is no compression done in the plugin when selecting a video on ios.

josiaho:
I agree - I don't see any compression in the plugin either. However, the video is definitely significantly compressed on import. Can the plugin be modified to force transcoding bypass?

PR #760, opened 2021-08-31
fix(ios): video quality

silvanadrian:
### Platforms affected
ios

### Motivation and Context

I use the `cordova-camera-plugin` to upload videos to youtube at which time I realized that literally every upload ends up on being `sd` only even though most smartphones these days are able to record much higher resolution videos.
**Before the change:**
Screenshot 2021-08-31 at 22 32 02
IMG_1556

**After the change:**
Screenshot 2021-08-31 at 22 21 53
(compression gets skipped, just shows up very shortly)

Closes https://github.com/apache/cordova-plugin-camera/issues/732

### Description

At the moment it's more of a quickfix than a proper solution, I guess I can even improve it further by adding a video quality option to the `CameraOptions` then everyone can choose themselfs if they would want to be able to get higher resolution videos.

Does anyone have an opinion on that?

For reference: https://stackoverflow.com/questions/20190485/how-to-avoid-compression-after-selecting-video-from-uiimagepickercontroller-in-i

### Testing

As described tested it by uploading the same video 2 times to youtube from a iPhone to make sure that a higher resolution video got uploaded which was the case.

### Checklist


Siedlerchr:
This fix works only for selecting existing videos from the photo library, not for taking new videos.

PR #775, opened 2021-11-09
fix(android): close cursors to suppress log warnings

manstie:
### Platforms affected

Android

### Motivation and Context

- "W/System: A resource failed to call close" is logged whenever a Cursor object is left unclosed in this file.
- `android.content.ContentResolver.query` returns:
```
A Cursor object, which is positioned before the first entry. May return null if the underlying content provider returns null, or if it crashes.
```
So we should check for `null` Cursors after running `queryImgDB`

### Description

- Added null checks before running functions on Cursor objects.
- Close all cursor objects after they have been created and used.

### Testing

I ran an app using this plugin in debug mode through Android Studio and confirmed the warnings no longer appear after these changes have been made.

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -287,7 +287,11 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
     public void takePicture(int returnType, int encodingType)
     {
         // Save the number of images currently on disk for later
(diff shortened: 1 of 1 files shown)


breautek:
Obsolete since #907 removed the usage involving `Cursor`.

Issue #779, opened 2021-11-19
Duplicate queries element in AppManifest.xml


Abrissirba:
# Bug Report

## Problem
Android SDK 30 requires you to define which other apps you app want to communicate with. This can be done by adding the following to config.xml

```
<config-file parent="queries" target="AndroidManifest.xml">
    <package android:name="<name of package>" />
</config-file>
```

### What is expected to happen?
When I run add platform I expect to get one queries element in AppManifest.xml

### What does actually happen?
I get two elements. One is the one I added myself in config.xml and the other is comming from plugin.xml in cordova-plugin-camera.

## Information
Introduced here: https://github.com/apache/cordova-plugin-camera/commit/abfbbd35d570f8bf60c3bcaa5a930cb74fdb5eac.

I have tried to manually edit the plugin.xml file locally to this and confirmed that it works

```
<config-file target="AndroidManifest.xml" parent="queries">
                <intent>
                    <action android:name="android.media.action.IMAGE_CAPTURE" />
                </intent>
                <intent>
                    <action android:name="android.intent.action.GET_CONTENT" />
                </intent>
                <intent>
                    <action android:name="android.intent.action.PICK" />
                </intent>
                <intent>
                    <action android:name="com.android.camera.action.CROP" />
                    <data android:scheme="content" android:mimeType="image/*"/>
                </intent>
        </config-file>
```` 

### Command or Code
cordova platform add android

### Version information
cordova-plugin-camera@6.0.0
"cordova-android": "^10.1.1",
10.0.0 (cordova-lib@10.1.0)

## Checklist

- [x ] I searched for existing GitHub issues
- [ x] I updated all Cordova tooling to most recent version
- [ x] I included all the necessary information above

PR #780, opened 2021-11-19
fix(android): update queries in plugin.xml

Abrissirba:
### Platforms affected
Android

### Motivation and Context
fixes https://github.com/apache/cordova-plugin-camera/issues/779

### Description
Change the queries element of plugin.xml so that not duplicate elements are created if other plugins or config.xml also contains a queries element

### Testing
Add the following to the projects config.xml and add android platform. Only one queries element should be created in AppManifest.xml.

<config-file parent="queries" target="AndroidManifest.xml">
    <package android:name="<name of package>" />
</config-file>

### Checklist

- [ x] I've run the tests to see all new and existing tests pass
- [x ] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -69,22 +69,20 @@
           </provider>
         </config-file>
 
-        <config-file target="AndroidManifest.xml" parent="/*">
-            <queries>
-                <intent>
-                    <action android:name="android.media.action.IMAGE_CAPTURE" />
-                </intent>
-                <intent>
-                    <action android:name="android.intent.action.GET_CONTENT" />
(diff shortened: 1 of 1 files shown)


erisu (approved):
I tested and confirmed the changes.

* The `queries` element is still created if missing.
* It removes the `intent` that this plugin defined when uninstalling.
* It does not remove other `intent` elements that were added by other plugins.

PR #810, opened 2022-09-28
fix(browser): use navigator.mediaDevices.getUserMedia

jcesarmobile:
navigator.getUserMedia has been deprecated for a long time, most browsers still support it, but others like safari already removed it and others probably will in the future.

this PR uses navigator.mediaDevices.getUserMedia if available.


erisu (approved):
LGTM

PR #814, opened 2022-10-13
feat(android)!: Android 13 support

ochakov:
### Platforms affected
Android

### Motivation and Context
Allow building applications using Android SDK 33, support for Android 13

Closes https://github.com/apache/cordova-plugin-camera/issues/825

### Description
Remove deprecated READ_EXTERNAL_STORAGE and WRITE_EXTERNAL_STORAGE permissions and replace with new READ_MEDIA_IMAGES 

### Testing

Taking pictures and accessing gallery, save pictures into album

### Checklist

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -56,6 +56,8 @@
         </config-file>
         <config-file target="AndroidManifest.xml" parent="/*">
             <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" android:maxSdkVersion="32" />
+			<uses-permission android:name="android.permission.READ_MEDIA_IMAGES" />
+			<uses-permission android:name="android.permission.READ_MEDIA_VIDEO" />
         </config-file>
         <config-file target="AndroidManifest.xml" parent="application">
           <provider
--- a/plugin.xml
+++ b/plugin.xml
@@ -55,7 +55,7 @@
             </feature>
         </config-file>
         <config-file target="AndroidManifest.xml" parent="/*">
-            <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+            <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" android:maxSdkVersion="32" />
         </config-file>
         <config-file target="AndroidManifest.xml" parent="application">
           <provider
--- a/plugin.xml
+++ b/plugin.xml
@@ -56,8 +56,8 @@
         </config-file>
         <config-file target="AndroidManifest.xml" parent="/*">
             <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" android:maxSdkVersion="32" />
-			<uses-permission android:name="android.permission.READ_MEDIA_IMAGES" />
-			<uses-permission android:name="android.permission.READ_MEDIA_VIDEO" />
+            <uses-permission android:name="android.permission.READ_MEDIA_IMAGES" />
+            <uses-permission android:name="android.permission.READ_MEDIA_VIDEO" />
         </config-file>
         <config-file target="AndroidManifest.xml" parent="application">
           <provider
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -277,7 +277,6 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
      */
     public void callTakePicture(int returnType, int encodingType) {
         String[] storagePermissions = getPermissions(true, mediaType);
-        boolean storagePermission = hasPermissions(storagePermissions);
         boolean saveAlbumPermission;
         if (this.saveToPhotoAlbum) {
             saveAlbumPermission = hasPermissions(storagePermissions);
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -122,7 +122,17 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
     private boolean orientationCorrected;   // Has the picture's orientation been corrected
     private boolean allowEdit;              // Should we allow the user to crop the image.
 
-    protected final static String[] permissions = { Manifest.permission.CAMERA, Manifest.permission.READ_EXTERNAL_STORAGE, Manifest.permission.WRITE_EXTERNAL_STORAGE };
+    protected static String[] permissions;
+    protected static String[] storagePermissions;
+    static {
+        if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+            permissions = new String[]{ Manifest.permission.CAMERA, Manifest.permission.READ_MEDIA_IMAGES };
+            storagePermissions = new String[]{ Manifest.permission.READ_MEDIA_IMAGES };
+        } else {
+            permissions = new String[]{ Manifest.permission.CAMERA, Manifest.permission.READ_EXTERNAL_STORAGE, Manifest.permission.WRITE_EXTERNAL_STORAGE };
+            storagePermissions = new String[]{ Manifest.permission.READ_EXTERNAL_STORAGE, Manifest.permission.WRITE_EXTERNAL_STORAGE };
+        }
+    }
 
     public CallbackContext callbackContext;
     private int numPics;
@@ -195,8 +205,8 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
                 }
                 else if ((this.srcType == PHOTOLIBRARY) || (this.srcType == SAVEDPHOTOALBUM)) {
                     // FIXME: Stop always requesting the permission
-                    if(!PermissionHelper.hasPermission(this, Manifest.permission.READ_EXTERNAL_STORAGE)) {
-                        PermissionHelper.requestPermission(this, SAVE_TO_ALBUM_SEC, Manifest.permission.READ_EXTERNAL_STORAGE);
+                    if(!hasPermissions(storagePermissions)) {
+                        PermissionHelper.requestPermissions(this, SAVE_TO_ALBUM_SEC, storagePermissions);
                     } else {
                         this.getImage(this.srcType, destType);
                     }
@@ -245,8 +255,12 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
      * @param encodingType           Compression quality hint (0-100: 0=low quality & high compression, 100=compress of max quality)
      */
     public void callTakePicture(int returnType, int encodingType) {
-        boolean saveAlbumPermission = PermissionHelper.hasPermission(this, Manifest.permission.READ_EXTERNAL_STORAGE)
-                && PermissionHelper.hasPermission(this, Manifest.permission.WRITE_EXTERNAL_STORAGE);
+        boolean saveAlbumPermission;
+        if (this.saveToPhotoAlbum) {
+            saveAlbumPermission = hasPermissions(storagePermissions);
+        } else {
+            saveAlbumPermission = true;
+        }
         boolean takePicturePermission = PermissionHelper.hasPermission(this, Manifest.permission.CAMERA);
 
         // CB-10120: The CAMERA permission does not need to be requested unless it is declared
@@ -277,8 +291,7 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
         } else if (saveAlbumPermission) {
             PermissionHelper.requestPermission(this, TAKE_PIC_SEC, Manifest.permission.CAMERA);
         } else if (takePicturePermission) {
-            PermissionHelper.requestPermissions(this, TAKE_PIC_SEC,
-                    new String[]{Manifest.permission.READ_EXTERNAL_STORAGE, Manifest.permission.WRITE_EXTERNAL_STORAGE});
+            PermissionHelper.requestPermissions(this, TAKE_PIC_SEC, storagePermissions);
         } else {
             PermissionHelper.requestPermissions(this, TAKE_PIC_SEC, permissions);
         }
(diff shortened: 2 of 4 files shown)


felicienfrancois:
@ochakov I tested your fork in my project and got 2 issues :
1°) At compilation time 
`Element uses-permission#android.permission.WRITE_EXTERNAL_STORAGE at AndroidManifest.xml:59:5-108 duplicated with element declared at AndroidManifest.xml:51:5-81`
It seems that manifest merger fail to merge
`<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" android:maxSdkVersion="32" />`
with
`<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />` from my other plugins.

=> removing  android:maxSdkVersion="32" solved the compilation issue.

2°) At runtime when trying to capture photo (either from camera or library), on Android 13, I always get permission refused.
After adding some logs in onRequestPermissionResult, I saw that the only permission requested is READ_MEDIA_IMAGES and the result is [-1] (PackageManager.PERMISSION_DENIED).

It seems that READ_MEDIA_IMAGES permission does not have to be requested.

=> Removing Manifest.permission.READ_MEDIA_IMAGES from requested permissions in CameraLauncher:129-130 solve the issue

breautek:
This is a breaking change as it requires at least apps compiled with API 33 and I'm sure the current tests are failing due to that reason since cordova-android@11 by default compiles with API 32.

So I created a new 7.0.0 milestone for this PR to be added to so this can be looked at after cordova-android is updated for API 33 support.

ochakov:
> It seems that READ_MEDIA_IMAGES permission does not have to be requested.

It does work for me, using Pixel 4 and 7 and it does not allow taking pictures if this permission is not granted. I think we must compare the parameters sent to the plugin when taking pictures. In my case, I am using destination type file without saving to album.
This permission might not be required when using destination type URI.

felicienfrancois:
> > It seems that READ_MEDIA_IMAGES permission does not have to be requested.
> 
> It does work for me, using Pixel 4 and 7 and it does not allow taking pictures if this permission is not granted. I think we must compare the parameters sent to the plugin when taking pictures. In my case, I am using destination type file without saving to album. This permission might not be required when using destination type URI.

On my side, i'm using destination type data URI on pixel 5

breautek:
> according to [Android docs](https://developer.android.com/reference/android/Manifest.permission#WRITE_EXTERNAL_STORAGE) , WRITE_EXTERNAL_STORAGE permission has no effect since Android 11 (API 30). The right code would be to set max sdk to 29 on this permission entry ?

Yes, it's actually been obsolete since scoped storage, which enabled in API 29, but can be opt out back to API 28 behaviour via `requestLegacyExternalStorage`. Starting in API 30, Scoped storage is forcefully enforced which makes the `WRITE_EXTERNAL_STORAGE` obsolete since API 30. However changing the manifest may have caveats on how it's handled in the code. For example... requesting permission even if it's a no-op is OK as long as it's declared, but attempting to request permission while not being declared I think will throw an error. So for sake of simplicity, I think we may want to keep it as maxSdkVersion `32` so that we can use the API 33 check as the PR stands now.

felicienfrancois:
Ok understood.
So I would not bet on other plugins making the exact same change with the exact same max sdk number.

fidaktk (approved):

jfoclpf:
Any time period for the reintegration of this PR? Thanks

jfoclpf:
> This is a breaking change as it requires at least apps compiled with API 33 and I'm sure the current tests are failing due to that reason since cordova-android@11 by default compiles with API 32.
> 
> So I created a new 7.0.0 milestone for this PR to be added to so this can be looked at after cordova-android is updated for API 33 support.

indeed, it is strange that the Android 12 test suite fails, but I can't see the test logs anymore

@ochakov can you kindly amend your commits such that it is Android 12 (SDK v 32) compatible? I.e., it passes the Test suite for Android 12.

breautek:
> Any time period for the reintegration of this PR? Thanks

It depends on cordova-android 12 which is unreleased and still in development.

> indeed, it is strange that the Android 12 test suite fails, but I can't see the test logs anymore
>
> @ochakov can you kindly amend your commits such that it is Android 12 (SDK v 32) compatible? I.e., it passes the Test suite for Android 12.

There's nothing to be done at this time. The code should work fine, it just needs to be compiled with SDK 33. The test fails because it uses the latest cordova-android version, which by default targets with SDK 32.

This PR should be usable as is if you set the `android-compileSdkVersion` to 33. You may also need to install build tools 33 and set `android-buildToolsVersion` to the full version name (e.g. `33.0.0`) if you want to test it yourself.

jfoclpf:
> This PR should be usable as is if you set the `android-compileSdkVersion` to 33. 

is this achievable merely by adding to config.xml this?

```xml
  <platform name="android">
    <preference name="android-compileSdkVersion" value="33"/>
    <preference name="android-targetSdkVersion" value="33"/>
```

jfoclpf:
> There's nothing to be done at this time. The code should work fine, it just needs to be compiled with SDK 33. The test fails because it uses the latest cordova-android version, which by default targets with SDK 32.

got it, thanks

breautek:
> > This PR should be usable as is if you set the `android-compileSdkVersion` to 33.
> 
> is this achievable merely by adding to config.xml this?
> 
> ```
>   <platform name="android">
>     <preference name="android-compileSdkVersion" value="33"/>
>     <preference name="android-targetSdkVersion" value="33"/>
> ```

Usually at cordova-android level, support just entails bumping build tools version to the next API level, and compiling for that target. Sometimes it may require more changes but in my experience that's usually all that is required.

`android-targetSdkVersion` can also be left at version 32 if you're not ready to deal with whatever changes API 33 brings.  This PR main requirement is that `android-compileSdkVersion` is set to 33 or later.

cordova-android@12 will have these values set to 33 by default.

keva91:
Hi , 
I tried to use the PR but i still can't build :/  : 
image


This is my setup : 
image
image
image

breautek:
@ochakov 

Have you tested this PR for taking videos? I noticed that right now we are only handling the `READ_MEDIA_IMAGES` permission, but do not handle the `READ_MEDIA_VIDEOS` permission which I suspect we do if we use the video API part of this plugin.

breautek:
> Hi , I tried to use the PR but i still can't build thinking I got this error : image
> 
> This is my setup : image image image

You have another plugin or routine that is adding a `WRITE_EXTERNAL_STORAGE` permission to your `AndroidManifest.xml` and it's creating a duplicate entry.

plugin.xml
```diff
@@ -55,7 +55,7 @@
             </feature>
         </config-file>
         <config-file target="AndroidManifest.xml" parent="/*">
-            <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+            <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" android:maxSdkVersion="32" />
         </config-file>
```
jfoclpf, line 59: I guess you should add here also 

```xml
<uses-permission android:name="android.permission.READ_MEDIA_IMAGES" android:minSdkVersion="33" />
```

based on suggestion from @felicienfrancois

jfoclpf: @ochakov could you add that to this PR, such that I can use this PR in my project?
thanks

ochakov: Made the following changes in this PR:
1. Video permission when taking video instead of image permission as requested by @breautek 
2. Updated manifest as requested by @jfoclpf

jfoclpf:
@ochakov thanks a lot for your amendments 

I just tested now and `WRITE_EXTERNAL_STORAGE` with `maxSdkVersion` creates a huge list of conflicts with other (also official, such as file) plugins. Other plugins don't mention `maxSdkVersion` for `WRITE_EXTERNAL_STORAGE` which creates a build failure. Please do kindly remove `maxSdkVersion` for `WRITE_EXTERNAL_STORAGE` (at least for now, maybe later `maxSdkVersion` can be added at same time for all official plugins).

On the other hand `READ_MEDIA_IMAGES` and `READ_MEDIA_VIDEO` are totally new permissions introduced only in API 33, thus do kindly add `android:minSdkVersion="33"` to these

Do kindly thus amend to the following:

```xml
  <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
  <uses-permission android:name="android.permission.READ_MEDIA_IMAGES" android:minSdkVersion="33" />
  <uses-permission android:name="android.permission.READ_MEDIA_VIDEO" android:minSdkVersion="33" />
```

I tested now with these amendments on Android 12 (API 32) and 13 (API 33) and it builds and works just fine.

plugin.xml
```diff
@@ -55,7 +55,9 @@
             </feature>
         </config-file>
         <config-file target="AndroidManifest.xml" parent="/*">
-            <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+            <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" android:maxSdkVersion="32" />
```
jfoclpf, line 58: As explained, do kindly amend here to:

```xml
  <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
  <uses-permission android:name="android.permission.READ_MEDIA_IMAGES" android:minSdkVersion="33" />
  <uses-permission android:name="android.permission.READ_MEDIA_VIDEO" android:minSdkVersion="33" />
```

I tested on both API 32 and 33 and it builds and works just fine.

ochakov:
@jfoclpf 
Is there any reason to put minSdkVersion for READ_MEDIA_IMAGES and READ_MEDIA_VIDEO permissions? I found it will compile for older targets without specifying minSdkVersion and I can't find any recommendation to use this tag.
Regarding maxSdkVersion for WRITE_EXTERNAL_STORAGE, this is a specific requirement from Android and it is listed in mandatory changes for apps targeting API 33. In our case, we had to change other plugins (BTW, I submitted a similar PR for cordova-plugin-media-capture containing maxSdkVersion) and include maxSdkVersion in all of them.
I suggest to hear other opinions in this thread if to leave maxSdkVersion as required or remove it for compatibility.

jfoclpf:
Hi @ochakov 

`WRITE_EXTERNAL_STORAGE` also compiles OK for API 33, (I just tested, @felicienfrancois and @keva91 also tested) it is simply ignored in API 33 and by avoiding maxSdk we avoid a lot of building conflicts.

By having minSdk we ensure these new permissions are only used with API 33 and above without any building conflicts, since these permissions are totally new with API 33.

@breautek do you have any opinion therefor?

felicienfrancois:
Maybe we should also raise an issue on the manifest merger. The conflict issue is caused my manifest merger not being able to merge 2 permissions having a different maxSdkVersion / minSdkVersion.
I think it could be improved to avoid such conflicts (taking the min / max value).

jfoclpf:
That's off topic but it's a very good point IMHO, to merge considering the
most restrictive feature, for example if there are two different maxSdk
select the lowest.

pradeep3093:
Any update on these feature?

breautek:
> Any update on these feature?

This PR is targeted for cordova-android@12 which is still in development.

jkainth:
Any updates to this PR?

breautek:
> Any updates to this PR?

Like the last few comments stated, this PR depends on cordova-android@12 which is in [development](https://github.com/apache/cordova-android/milestone/13).

This PR is blocked until cordova-android@12 is released, and only then this PR will resume (cordova-android@12 should fix the failing tests).

Locking to keep the discussion history clean. Will unlock once cordova-android@12 is released to allow for further comments.

breautek:
Closing/reopening to restart tests.

src/android/CameraLauncher.java
```diff
@@ -223,6 +223,37 @@ else if ((this.srcType == PHOTOLIBRARY) || (this.srcType == SAVEDPHOTOALBUM)) {
...
+                    case VIDEO:
+                        return new String[]{ Manifest.permission.READ_MEDIA_VIDEO };
+                    default:
+                        return new String[]{ Manifest.permission.READ_MEDIA_IMAGES, Manifest.permission.READ_MEDIA_VIDEO };
+                }
+            }
+            else {
+                switch (mediaType) {
+                    case PICTURE:
+                        return new String[]{ Manifest.permission.CAMERA, Manifest.permission.CAMERA, Manifest.permission.READ_MEDIA_IMAGES };
```
Kohagan91, line 241:
```suggestion
                        return new String[]{ Manifest.permission.CAMERA, Manifest.permission.READ_MEDIA_IMAGES };
```
Duplicate permission ?

epetre:
This PR works well on my Android s20 FE.
The only issue remaining is the Gallery, for some reason, it just complains about permissions being denied when trying to access the gallery.

src/android/CameraLauncher.java
```diff
@@ -245,8 +276,14 @@ private String getTempDirectoryPath() {
...
     public void callTakePicture(int returnType, int encodingType) {
-        boolean saveAlbumPermission = PermissionHelper.hasPermission(this, Manifest.permission.READ_EXTERNAL_STORAGE)
-                && PermissionHelper.hasPermission(this, Manifest.permission.WRITE_EXTERNAL_STORAGE);
+        String[] storagePermissions = getPermissions(true, mediaType);
+        boolean saveAlbumPermission;
+        if (this.saveToPhotoAlbum) {
+            saveAlbumPermission = hasPermissions(storagePermissions);
+        } else {
+            saveAlbumPermission = true;
+        }
```
DaedalusDev, line 285: Will produce and issue with Android < 8.0. I've replace the code above with this and it works.

```diff
        boolean saveAlbumPermission = hasPermissions(storagePermissions);
```

Kaabi:
is keep add duplicate writes to external into manifest I think it conflict with Android permissions plugin

```
> Task :app:processDebugMainManifest FAILED
~platforms\android\app\src\main\AndroidManifest.xml:26:5-108 Error:                                                                                                               
        Element uses-permission#android.permission.WRITE_EXTERNAL_STORAGE at AndroidManifest.xml:26:5-108 duplicated with element declared at AndroidManifest.xml:25:5-81
~platforms\android\app\src\main\AndroidManifest.xml Error:
        Validation failed, exiting
```

ndrake:
In our testing, this PR works for us with Cordova Android 12

Issue #744, opened 2021-07-04
App crash after allowing the camera permission


Mohammad-Toufiq-Pathan:
Hi all, I am using Android 11  camera plugin 5.0.2, when I click on camera icon permission dialog will appear when I allow the permission app getting crash, this is happening only in release build, in debug build it's working fine, please help me out

dantheman2865:
I am also experiencing the same problem on 5.0.3. Here are the exception logs from my Android App:

```
2021-08-10 08:31:27.551 4303-4598/my.app.name V/Capacitor/Plugin: To native (Cordova plugin): callbackId: Camera770925669, service: Camera, action: takePicture, actionArgs: [100,0,1,-1,-1,0,0,false,false,false,null,0]
2021-08-10 08:31:27.563 4303-4598/my.app.name W/System.err: java.lang.ClassNotFoundException: my.app.name.BuildConfig
2021-08-10 08:31:27.657 4303-4598/my.app.name W/System.err:     at java.lang.Class.classForName(Native Method)
2021-08-10 08:31:27.657 4303-4598/my.app.name W/System.err:     at java.lang.Class.forName(Class.java:454)
2021-08-10 08:31:27.657 4303-4598/my.app.name W/System.err:     at java.lang.Class.forName(Class.java:379)
2021-08-10 08:31:27.658 4303-4598/my.app.name W/System.err:     at org.apache.cordova.BuildHelper.getBuildConfigValue(BuildHelper.java:54)
2021-08-10 08:31:27.659 4303-4598/my.app.name W/System.err:     at org.apache.cordova.camera.CameraLauncher.execute(CameraLauncher.java:149)
2021-08-10 08:31:27.659 4303-4598/my.app.name W/System.err:     at org.apache.cordova.CordovaPlugin.execute(CordovaPlugin.java:98)
2021-08-10 08:31:27.660 4303-4598/my.app.name W/System.err:     at org.apache.cordova.PluginManager.exec(PluginManager.java:132)
2021-08-10 08:31:27.660 4303-4598/my.app.name W/System.err:     at com.getcapacitor.MessageHandler.callCordovaPluginMethod(MessageHandler.java:123)
2021-08-10 08:31:27.661 4303-4598/my.app.name W/System.err:     at com.getcapacitor.MessageHandler.postMessage(MessageHandler.java:61)
[...]

PR #827, opened 2023-03-14
fix: set applicationId

seamlink-aalves:
### Description
Fix camera launcher https://github.com/apache/cordova-plugin-camera/issues/744

### Testing
I invoked the camera plugin successfully

### Checklist

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -147,7 +147,7 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
         this.callbackContext = callbackContext;
         //Adding an API to CoreAndroid to get the BuildConfigValue
         //This allows us to not make this a breaking change to embedding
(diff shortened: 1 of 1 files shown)


breautek (approved):
I think this is a good overall change.

[org.apache.cordova.BuildHelper](https://github.com/apache/cordova-android/blob/master/framework/src/org/apache/cordova/BuildHelper.java) is a utility class for accessing the app's `BuildConfig` class, which is a class that is generated by gradle builds. It won't exist if the project happens.

Additionally the `BuildHelper` class makes use of reflections, which can make debugging difficult and stacktraces mostly useless (since it will hard what you're trying to invoke).

[Context.getPackageName](https://developer.android.com/reference/android/content/Context#getPackageName()) is a SDK level 1 API which returns the same information as the OS knows it (powered by the AndroidManifest). It should be guaranteed to be accurate and available.

While the `BuildHelper` class may have uses, if we can get the same information from concrete android APIs, I think that should be the more preferable approach :+1:

erisu (approved):

PR #828, opened 2023-03-16
Illegal Argument Exception with Ionic 6 | applicationId is null issue

ImAbhishekTomar:
### Checklist


breautek:
I'm closing this because I don't think it's applicable anymore.

We no longer rely on BuildConfig for the application ID and the PR code is very similar to what exists in the current code base.

PR #835, opened 2023-06-04
ci: Drop API 22 & 31. Added API 24 & 33

breautek:
### Platforms affected

CI

Drops API 22 since that is no longer supported by cordova-android.
Drops API 31 because API 32 is a feature update on top of API 31, they are nearly identical API levels.

Added API 24 since that is our new minimum target.
Added API 33 for completeness, especially since this plugin requires changes for API 33 support.

### Checklist


erisu (changes requested):
Can it be synced with paramedic's workflow?

I see the android-emulator-runnner dependency was not updated. I think that is the only missing change but could try a full sync to see.

breautek:
> Can it be synced with paramedic's workflow?
> 
> I see the android-emulator-runnner dependency was not updated. I think that is the only missing change but could try a full sync to see.

I copied the entire paramedic version over, git diff shows only the reactivecircus runners have updated.

erisu (approved):

PR #868, opened 2023-11-08
fix(android): Fixed exif handling when loading picture from gallery

### LINKED ISSUE #867: Exif data lost if picture retrieved from gallery in DATA_URL on android
# Bug Report

### What is expected to happen?

Have the base64 of the picture with all the original jpeg exif metadata in a data url

### What does actually happen?

The given picture in the data url never see its exif data back

## Information

- Load an image from the gallery (i use jpeg) on android
- Check the exif data

Also, I looked at the code, and there should be a loading of exif data in `android/CameraLauncher.java:1300` in the `processPicture` method or at least before the execution of `processPicture` on `android/CameraLauncher.java:781`.

### Command or Code

```
// This is approximate since i work on angular with @awesome-cordova-plugins/camera
getPicture({
        quality: 50,
        destinationType: DestinationType.DATA_URL,
        encodingType: EncodingType.JPEG,
        mediaType: MediaType.PICTURE,
        targetHeight: 720,
        targetWidth: 1280,
        sourceType: PictureSourceType.PHOTOLIBRARY,
    });
```

### Environment, Platform, Device

Galaxy Tab S8, Android 13

### Version information

Ionic 5,
Android 13,
Editing on windows,
Angular 10

## Checklist

- [x] I searched for existing GitHub issues
- [ ] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

mariusvn:
### Platforms affected

- Android

### Motivation and Context

The exif data was removed if we loaded from the gallery in DATA_URL format.

This should fix https://github.com/apache/cordova-plugin-camera/issues/867.

### Description

Instead for returning the bitmap compressed and encoded in base64, we save it into a file, and apply the input exif data and then we put it back in the memory to encode it in base64 and return it.

### Testing

I tested my changes in an ionic app that use this feature via `@awesome-cordova-plugins/camera@5.46.0`

### Checklist

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -783,9 +783,17 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
                         try {
                             String modifiedPath = this.outputModifiedBitmap(bitmap, uri, mimeTypeOfGalleryFile);
                             InputStream fileStream = FileHelper.getInputStreamFromUriString(modifiedPath, cordova);
-                            byte[] file = new byte[fileStream.available()];
+                            ByteArrayOutputStream buffer = new ByteArrayOutputStream();
 
-                            fileStream.read(file);
+                            int nRead;
+                            byte[] data = new byte[512];
+
+                            while ((nRead = fileStream.read(data, 0, data.length)) != -1) {
+                                buffer.write(data, 0, nRead);
+                            }
+
+                            buffer.flush();
+                            byte[] file = buffer.toByteArray();
 
                             byte[] output = Base64.encode(file, Base64.NO_WRAP);
                             String js_out = new String(output);
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -778,7 +778,22 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
 
                     // If sending base64 image back
                     if (destType == DATA_URL) {
-                        this.processPicture(bitmap, this.encodingType);
+                        // To allow keeping the Exif data, we need to save the image to allow ExifInterface add the data
+                        // After that, we load the image and send it as base64
+                        try {
+                            String modifiedPath = this.outputModifiedBitmap(bitmap, uri, mimeTypeOfGalleryFile);
+                            InputStream fileStream = FileHelper.getInputStreamFromUriString(modifiedPath, cordova);
+                            byte[] file = new byte[fileStream.available()];
+
+                            fileStream.read(file);
+
+                            byte[] output = Base64.encode(file, Base64.NO_WRAP);
+                            String js_out = new String(output);
+                            this.callbackContext.success(js_out);
+                        } catch (IOException e) {
+                            e.printStackTrace();
+                            this.failPicture("Error retrieving image: "+e.getLocalizedMessage());
+                        }
                     }
 
                     // If sending filename back
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -786,7 +786,9 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
                             ByteArrayOutputStream buffer = new ByteArrayOutputStream();
 
                             int nRead;
-                            byte[] data = new byte[512];
+                            final int MB = 1048576;
+
+                            byte[] data = new byte[MB * 4];
 
                             while ((nRead = fileStream.read(data, 0, data.length)) != -1) {
                                 buffer.write(data, 0, nRead);


src/android/CameraLauncher.java
```diff
@@ -778,7 +778,22 @@ private void processResultFromGallery(int destType, Intent intent) {
 
                     // If sending base64 image back
                     if (destType == DATA_URL) {
-                        this.processPicture(bitmap, this.encodingType);
+                        // To allow keeping the Exif data, we need to save the image to allow ExifInterface add the data
+                        // After that, we load the image and send it as base64
+                        try {
+                            String modifiedPath = this.outputModifiedBitmap(bitmap, uri, mimeTypeOfGalleryFile);
+                            InputStream fileStream = FileHelper.getInputStreamFromUriString(modifiedPath, cordova);
+                            byte[] file = new byte[fileStream.available()];
```
breautek, line 786: Using `available()` here is unsafe. `available` API is used to determine how many bytes are available to read without blocking the current thread. It may return `0` or any number that is less than the actual size of the file. This can lead to file corruption since the byte buffer is used to attempt to read the entire file. See https://developer.android.com/reference/java/io/InputStream#available()

Instead, create a chunk sized buffer and use the [read](https://developer.android.com/reference/java/io/InputStream#read(byte[],%20int,%20int)) API in a loop until `read` returns `-1` which tells you that you read the entire file. Note that the whole buffer may not be filled, so using the return value to know how many bytes were read is required.

breautek (changes requested):

src/android/CameraLauncher.java
mariusvn: working on it !

mariusvn:
should be done

src/android/CameraLauncher.java
```diff
@@ -778,7 +778,30 @@ private void processResultFromGallery(int destType, Intent intent) {
...
-                        this.processPicture(bitmap, this.encodingType);
+                        // To allow keeping the Exif data, we need to save the image to allow ExifInterface add the data
+                        // After that, we load the image and send it as base64
+                        try {
+                            String modifiedPath = this.outputModifiedBitmap(bitmap, uri, mimeTypeOfGalleryFile);
+                            InputStream fileStream = FileHelper.getInputStreamFromUriString(modifiedPath, cordova);
+                            ByteArrayOutputStream buffer = new ByteArrayOutputStream();
+
+                            int nRead;
+                            byte[] data = new byte[512];
```
breautek, line 789: Looks good, I enabled the unit tests but I think we probably want to use a larger buffer than `512` bytes here.

default memory limits of android apps are `256mb` and this is used for images which is probably on average 5-10mb depending on the resolution of the captured image.

My Galaxy Note 9 takes pictures and it ranges from 3-5mb in JPG format.

```suggestion
                            final int MB = 1048576;
                            byte[] data = new byte[MB * 4];
```

Some phones may have larger images, other phones may have smaller images, but here we have a 4mb buffer which isn't too much compared to the 256mb memory limit of the app, and devices today aren't so memory constrained that a 4mb buffer will matter. The 4mb buffer will likely be large enough to fit an entire image, which is the closest behaviour that you had previously when you were trying to use `.avaiable()`, allowing the file to be read copied in a single loop pass.

If the image happens to be larger, then we drastically reduce the amount of times it needs to reset and fill the buffer. (e.g. if the image was 5mb, then we would need to do `10240` iterations to read the entire image with a 512 byte buffer, which would be kinda slow).

The `MB` constant is used here purely for readability so we know the value is 1 megabyte.

breautek:
FYI we can ignore the failing Chrome test suite.

GitToTheHub:
Hi can you fix the merge conflicts? Than it can be merged.

GitToTheHub:
I think it's not ready to merge. Claude Sonnet 4.6 tells:

## What it does

When loading a gallery image in DATA_URL format, instead of encoding the decoded bitmap directly (which loses EXIF), it saves to a temp file via outputModifiedBitmap (which copies EXIF), then reads it back and encodes to base64.

## Issues to address before merging

- Resource leaks — fileStream and buffer are never closed. Should use try-with-resources:
```java
try (InputStream fileStream = ...; ByteArrayOutputStream buffer = ...) { ... }
```
- Excessive memory allocation — new byte[MB * 4] allocates 4 MB per call. A standard 8–16 KB buffer is idiomatic.
- Temp file not cleaned up — outputModifiedBitmap writes a file that is never deleted after reading back.
- e.printStackTrace() — Should use the existing LOG.e(TAG, ...) pattern used throughout this codebase.
- Style — js_out uses snake_case in Java; jsOut is conventional
- JPEG re-compression — Saving the bitmap and reading it back applies another round of lossy JPEG compression, degrading image quality. The EXIF could be applied directly from the original uri without re-encoding the bitmap.
- Checklist — No tests run or added; the issue link uses "should fix" which won't auto-close #867 (needs Fixes #867 or Closes #867).

mariusvn:
Hello. Well, Its been 3 years, i had the time to work two years in another company and leave it so I won't work on this PR anymore. If anyone want to fix it, you're more than welcome to do so but sadly, i'm not working with this tool anymore.

GitToTheHub:
No problem, thanks for your reply :)

Issue #875, opened 2023-12-09
Error Code 20 on Android with Plugin v7.0.0


scyclops:
# Bug Report

## Problem

I'm seeing error code 20 returned as a result of calls to navigator.camera.getPicture on Android.  It's happening pretty regularly with between 2 and 15 reports a day.  I'm using cordova-plugin-camera version 7.0.0 with cordova-android 12.0.0

### What is expected to happen?

User should be able to use their device camera to take a photo and the image data should be returned to the getPicture success callback.

### What does actually happen?

The getPicture error callback is called with error code 20.

## Information

This issue was previously reported in #826

I had previously tried using the [jalios fork](https://github.com/jalios/cordova-plugin-camera) to fix this problem before cordova-plugin-camera version 7.0.0 was released but the jalios fork also resulted in error code 20 reports.

### Code

```javascript
// after deviceready event

let Camera = window.Camera,
    options = {
        quality: 75,
        destinationType: Camera.DestinationType.FILE_URI,
        sourceType: Camera.PictureSourceType.CAMERA,
        allowEdit: false,
        encodingType: Camera.EncodingType.JPEG,
        targetWidth: 1024,
        targetHeight: 1024,
        correctOrientation: true
    };

navigator.camera.getPicture(
    () => {
        console.log('success');
    },
    (error) => {
        console.log('error: ' + JSON.stringify(error));
    },
    options
);
```

### Environment, Platform, Device

On Android devices using cordova-plugin-camera version 7.0.0 and cordova-android 12.0.0

The error has occurred with the following devices (this is not an exhaustive list):

-- Android 12 --
Nokia 5.4
ITEL A60
Samsung Galaxy A21s
Oppo A16s
Moto G Stylus 5G

-- Android 11 --
Samsung Galaxy A50
Samsung Galaxy A10

-- Android 10 --
Moto G7 Plus
Moto G7 Power
Moto G7 Play
Samsung Galaxy Note9
Samsung Galaxy S9+

-- Android 8.1.0 --
Samsung Galaxy J5

fjckls:
Same here on devices with Android 13, getting the same **error:20**.

fjckls:
OK, I think I found it. Looking at the source code inside **_CameraLauncher.java_** we have `public static final int PERMISSION_DENIED_ERROR = 20;`. So I did a little investigation and stumbled upon these 2 permissions that are used in this plugin:

[...]

Issue #783, opened 2022-01-12
camera.getPicture() with FILE_URI destination returns invalid/unexpected image URI


wencywww:
# Bug Report

### What is expected to happen?

camera.getPicture() to return a valid file_URI in the form of `content://media/external/images/media/2` when Camera.DestinationType = FILE_URI

### What does actually happen?
the returned value is something like: `file///data/user/0/com.tld.appname/cache/1642011143203.jpg` and this seems to be an unreacheable url for the webview (can not render the picture). Moreover, searching through the filesystem i can't see such a file. There is no problems when the Camera.DestinationType = DATA_URL, however it is undesired due to possible memory problems

### Command or Code

`navigator.camera.getPicture(onSuccess, onFail, {
            quality: 50,  destinationType: Camera.DestinationType.FILE_URI
        });`

### Environment, Platform, Device

Developing on Windows 10, the result is with Android 7.0 phone

### Version information

cordova@11.0.0
cordova-android@10.1.1
cordova-plugin-camera@6.0.0
Android SDK Platform 11.0 / API Level 30
Gradle 7.3.3

## Checklist

-  I searched for existing GitHub issues

-  I updated all Cordova tooling to most recent version
- [ ] I included all the necessary information above

cmarian5:
I have the same issue. Did you found some solution?

wencywww:
> 

I don't think so :(
Asked within StackOverflow here
[https://stackoverflow.com/questions/70693557/camera-getpicture-returns-an-unexpected-file-uri-on-android](https://stackoverflow.com/questions/70693557/camera-getpicture-returns-an-unexpected-file-uri-on-android)
but with little success.
One suggested using this:
`<preference name="AndroidInsecureFileModeEnabled" value="true" />`
but the resulting url is still like `file//`




> I have the same issue. Did you found some solution?

kirkHawksworth:
Same issue

cmarian5:
For me <preference name="AndroidInsecureFileModeEnabled" value="true" /> worked fine. Meaning I could obtain a valid url of the image to add it in a html img tag.

kirkHawksworth:
@cmarian5  do you still receive the image url like: file:///data/user/0/package.name/cache/1645202153127.jpg

I have that preference enabled in my config.xml, added & removed Android etc. and still get the same issue

cmarian5:
[...]

Issue #761, opened 2021-09-06
Not allowed to load local resource 


clarklight:
Not displaying Photo after success call back 
Not allowed to load local resource

### What is expected to happen?
After taking a photo, using the file path to display the photo, by setting it into the src tag.

### What does actually happen?
The image display is broken and the console is showing, the only method i can display a photo right now, is to use the base64 DATA_URL method, but i would like to displaying using the file path. 
Not allowed to load local resource: 
The file path is as such:
file:///data/user/0/com.abc123.abc123app/cache/1630904187041.jpg

### Command or Code
navigator.camera.getPicture(onPhotoDataSuccess, onFail, {
    quality: 100,
    targetWidth: 1000,
    targetHeight: 1330,
    destinationType: Camera.DestinationType.NATIVE_URI,
    saveToPhotoAlbum: true,
    correctOrientation: true 
});
   // also tested with FILE_URI, and all possible setting for destinationType, and it does return the correct file path, but the file path isnt allowed to be displayed....and showing the not allowed to load local resource error...does it have something to do with Google changing the location for secure storage?

   function onPhotoDataSuccess(imageURI) {
      console.log(imageURI);
     var cameraImage = document.getElementById('image');
     cameraImage.src = imageURI;
}

function onFail(message) {
    console.log(message);
}

Building on:
Cordova Android 10.0.1
Camera plugin version 6.0.0
Testing on:
Samsung s10e

Thanks
Clark

alexandremsouza1:
I have the same problem

clarklight:
> I have the same problem

Do we know if it has anything to do with the google changing secure storage location? 
I have no idea what is the causes for this.

veronicatc:
I have the same issue. I use destinationType = FILE_URI.

The camera plugin is returning this for example for a photo selected from Downloads folder:

content://com.android.providers.downloads.documents/document/msf%3A48

And I get the error "Not allowed to access local resource" when setting it in the src of an IMG tag.

m30102:
use Base64-encoded 

navigator.camera.getPicture(onSuccess, onFail, { quality: 25,
    destinationType: Camera.DestinationType.DATA_URL
});

function onSuccess(imageData) {
    var image = document.getElementById('myImage');
    image.src = "data:image/jpeg;base64," + imageData;
}

JonSmart:
[...]

Issue #746, opened 2021-07-14
Android: FileURL returned by getPicture with sourceType photo album cannot be resolved


dominic-simplan:
# Bug Report

### What is expected to happen?
Calling `camera.getPicture()` with options
```
getPicture({
   sourceType: Camera.PictureSourceType.SAVEDPHOTOALBUM,
   destinationType: Camera.DestinationType.FILE_URI,
}
```
should return a fileURL which can be processed with `window.resolveLocalFileSystemURL()`

### What does actually happen?
If I select an image from my photo gallery, getPicture returns "`/storage/emulated/0/DCIM/Camera/PXL_xyz.jpg`".
When I try to resolve this URL with `window.resolveLocalFileSystemURL`, I get `FileError {code: 5}`.

Only files which [are modified by the plugin](https://github.com/apache/cordova-plugin-camera/blob/d356030070d903a929b7fb2adae7b9b53ab3dc7a/src/android/CameraLauncher.java#L759) seem to work. For example, if I select a png file but the (default) encodingType is set to JPEG, I get the following URL: "`file:///data/user/0/de.appname/cache/Screenshot_20210607-075901.png?1626272750921`". This can be resolved correctly.

## Information

Already tried adding the following to my config.xml but nothing changed:
```
<edit-config file="app/src/main/AndroidManifest.xml" mode="merge" target="/manifest/application">
      <application android:requestLegacyExternalStorage="true" />
</edit-config>
```

### Environment, Platform, Device
Google Pixel 4a
Android 11

### Version information
Cordova CLI 10.0.0
cordova-android: 9.0.0
cordova-plugin-camera: 5.0.1
cordova-plugin-file: 6.0.2

## Checklist

- [x] I searched for existing GitHub issues
- [ ] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

bjdose:
Same problem here. I've been trying to fix it without success.

skout90:
+1 same problem

robman70:
Same problem here. No solution?

rtherry:
+ Same problem here. With the latest Cordova stack 12.0.0 cordova-plugin-camera 7.0.0

PR #902, opened 2024-10-25
fix(android): always return uris

breautek:
This PR contains https://github.com/apache/cordova-plugin-camera/pull/901, will rebase after it's merged.

### Platforms affected

Android

### Motivation and Context

closes https://github.com/apache/cordova-plugin-camera/issues/875
closes https://github.com/apache/cordova-plugin-camera/issues/783
closes https://github.com/apache/cordova-plugin-camera/issues/761
closes https://github.com/apache/cordova-plugin-camera/issues/422
closes https://github.com/apache/cordova-plugin-camera/issues/746
closes https://github.com/apache/cordova-plugin-camera/issues/347

The `finalLocation` returns raw file paths, which isn't always usable especially with Scoped Access Framework. Instead return the stringified URI, which has temporary permission granted.

With file plugin 8.1.1 `content://` paths are resolvable.

**NOTE**: I've since realised that `DATA_URL` also doesn't properly return the data url (it is missing the scheme prefix). I however will consider that out of scope and correct that in another PR.

### Description

Removed finding the raw file path, which was not used other than to return it.

All `finalLocation` usage, which was used to return back to the webview now returns `uriString` instead. This usually a `file://` or a `content://` uri.

Simplified logic and removed conditions surrounding handling `finalLocation`.

### Testing

Manual testing images picked from gallery of different formats.
Paramedic tests passes.

### Checklist

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -728,77 +728,68 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
             }
         }
 
-        String fileLocation = FileHelper.getRealPath(uri, this.cordova);
-        LOG.d(LOG_TAG, "File location is: " + fileLocation);
-
         String uriString = uri.toString();
-        String finalLocation = fileLocation != null ? fileLocation : uriString;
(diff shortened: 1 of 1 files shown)


erisu (approved):
LGTM

PR #903, opened 2024-10-25
fix(android): Improper serialization of image uri in save instance state

breautek:
### Platforms affected

Android

### Motivation and Context

The imageUri field is not properly restored if the activity is killed,
which can cause further issues.

This is because `imageFilePath` was used instead for the imageUri which only contains the
path of the original URI, losing the scheme in the process.

### Description

Serialize using `imageUri.toString()` instead of using the imageFilePath

### Testing

paramedic testing + testing with `Don't Keep Activities` developer option enabled.

### Checklist

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -1375,7 +1375,7 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
         }
 
         if (this.imageUri != null) {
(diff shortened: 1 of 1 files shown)


breautek:
For clarity... The difference is `imageUri` will start off with a Uri object representing `content://some-path/...` or `file:///some-path/...` but if the activity was destroyed, it will get restored with an Uri representing `/some-path/`, missing the Uri scheme.

erisu (approved):

PR #909, opened 2024-10-25
refactor(android): Make WRITE_EXTERNAL_STORAGE optional

breautek:
### Platforms affected

Android

### Motivation and Context

Depends on https://github.com/apache/cordova-plugin-camera/pull/907

closes https://github.com/apache/cordova-plugin-camera/issues/861
closes https://github.com/apache/cordova-plugin-camera/issues/876

Across all API levels that we support, no permissions are necessary for reading from
gallery. This is because we use the Media Store and gain temporary access to the content.

Across all API levels that we support, no permissions are necessary for capturing an image from the camera. This is because we use an internal app cache directory and we grant the camera app to use this path.

Where permissions might be necessary is with the `saveToPhotoAlbum` option. If enabled, non scoped access devices (API 24-28) do require `WRITE_EXTERNAL_STORAGE` to write to the device's gallery. API 29+ with scoped access framework, no permissions are required.

This PR allows to be more modern on modern devices, and allows the user to opt into `WRITE_EXTERNAL_STORAGE` by providing the necessary `config.xml` directives, if they plan on using `saveToPhotoAlbum` option.

Users that don't want to save to the photo album does not need to declare any extra permissions.

### Description

Reworked permission management so that `WRITE_EXTERNAL_STORAGE` is optional. `READ_EXTERNAL_STORAGE` is completely obsolete and has been removed.

The `CAMERA` permission workaround is still relevant across all API levels. If declared, the permission must be granted to use the camera intent and that workaround still remains.

### Testing

Manual testing on API 28 & 34.
Paramedic test passes

### Checklist

- [x] I've run the tests to see all new and existing tests pass
- [x] I added automated test coverage as appropriate for this change
- [x] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [x] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [x] I've updated the documentation if necessary

PR #913, opened 2024-10-26
docs: Revisions for v8 public API changes with the return string form…

breautek:
…ats of getPicture

### Platforms affected

Documentation

### Motivation and Context

Added notes mostly in regards to the public API changes for the return format of `getPicture`.
Also improved documentation surrounding best practices.

### Description

- Added notes about return formats in v8+
- Added notes about the return formats used in v7 and earlier
- Removed `createFileEntry` example, doesn't seem relevant to the camera plugin anymore.
- Added `copyTo` example, as should be a common pattern starting in v8 moving forward.
- Added upload example, using standard browser APIs.
- Added example of how to render content when data uris used.
- Added example of how to render content when file uri is used, via `fileEntry.toURL()`
- Added notes indicating that returned file uris are temporary access grants only that typically only survive until the app is closed. For camera captures, this isn't completely true as app cache directory are used and it will survive much longer, but I feel it is easier to treat it as short-lived temporary for the purposes of illustrating the concept in general. (`cleanup` should be used anyway which will clean up these temporary files)
- Added notes stating that the `FILE_URI` returns is an implementation detail, and they should not assume it will always be `content://` or `file://`, etc... This is especially true for android devices.

### Testing

VSCode markdown preview

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,8 +135,31 @@ To add these entries into the `info.plist`, you can use the `edit-config` tag in
 
 ### camera.getPicture(successCallback, errorCallback, options)
 Takes a photo using the camera, or retrieves a photo from the device's
-image gallery.  The image is passed to the success callback as a
-Base64-encoded `String`, or as the URI for the image file.
+image gallery. The result is provided in the first parameter of the `successCallback` as a string.
+
+As of v8.0.0, the result is formatted as URIs. The scheme will vary depending on settings and platform.
+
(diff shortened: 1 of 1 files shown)


breautek:
Note that additional documentation will be added from https://github.com/apache/cordova-plugin-camera/pull/909 for making `WRITE_EXTERNAL_STORAGE` optional.

erisu (approved):

PR #914, opened 2024-10-28
deprecation: allowEdit

breautek:
### Platforms affected

All

### Motivation and Context

Deprecate allowEdit. It's already known and documented that on Android that it's unreliable and unsafe to use for reasons outside of our control. While it might be stable enough for iOS, I don't think we can ever provide the same capability for Android without grossly entering scope creep for the plugin responsibilities.

Generally speaking we should refrain from having platform-specific features. Therefore I propose to formally deprecate it in v8.0.0 so that we can indicate that we don't intend to support issues arising form using `allowEdit`, either on iOS or on android.

Note: While I have language indicating that it will be removed v9.0.0, that will obviously depend on who decides to make that release in the future, so don't consider have to consider it as a promise, but it should serve as a sufficient warning that in v9 or any future release the feature can disappear.

### Description

Adds a deprecation warning in the JS console, if `getPicture` is used with `allowEdit === true`

### Testing

`npm test`

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,7 +257,7 @@ Optional parameters to customize the camera settings.
 | quality | <code>number</code> | <code>50</code> | Quality of the saved image, expressed as a range of 0-100, where 100 is typically full resolution with no loss from file compression. (Note that information about the camera's resolution is unavailable.) |
 | destinationType | <code>[DestinationType](#module_Camera.DestinationType)</code> | <code>FILE_URI</code> | Choose the format of the return value. |
 | sourceType | <code>[PictureSourceType](#module_Camera.PictureSourceType)</code> | <code>CAMERA</code> | Set the source of the picture. |
-| allowEdit | <code>Boolean</code> | <code>false</code> | Allow simple editing of image before selection. |
+| ~~allowEdit~~ | <code>Boolean</code> | <code>false</code> | **Deprecated**. Allow simple editing of image before selection. |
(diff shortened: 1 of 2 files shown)


erisu (approved):
+1 for deprecating `allowEdit`.

Since this feature is unreliable and unsafe on Android, it would be better to remove it. Removing it from only one platform and making it platform-specific also seems like a poor approach, as we generally aim to keep our core plugins cross-platform compatible.

One change request would be to revise the wording from 'next major release' to 'a future release' to allow flexibility for removing it even beyond 9.x, in case we miss the intended target or need to delay the deprecation.

We might need to submit a discussion thread on the mailing list before completing the removal.

However, since this PR only adds a deprecation warning without actually removing the feature, we might be able to reach a consensus within this PR to include it in version 8.0.0. If the mailing list discussion ultimately decides not to remove it, we can easily revert the deprecation warning in a patch release.

jcesarmobile (approved):
Even on iOS the feature is confusing because it creates a square to crop the image that most people want to avoid and that doesn’t work very well anyway when zooming the image.

breautek:
> However, since this PR only adds a deprecation warning without actually removing the feature, we might be able to reach a consensus within this PR to include it in version 8.0.0. If the mailing list discussion ultimately decides not to remove it, we can easily revert the deprecation warning in a patch release.

I'll send a proposal thread in the mailing list for a call out. I'm also going to initiate the release buy-in and I'll explicitly mention this open PR as well, stating that the release may or may not include the PR depending on the consensus reached by tomorrow.

NiklasMerz (approved):

PR #950, opened 2026-01-30
ios: rename variables `cameraPicker` to `cdvUIImagePickerController`

GitToTheHub:
- Make clear, that `CDVUIImagePickerController` is used for these variables

### Checklist


dpogue (commented):
This feels a bit like refactoring for the sake of refactoring and making the code harder to read by using more complex variable names that don't indicate their purpose.

In particular, `weakCDVUIImagePickerController` is a weird name to use for a variable. It doesn't matter if it's a weak reference or not, it works like any other variable. The only reason `weakSelf` is a convention is because you need to explicitly create a variable to hold the weak reference.

PR #951, opened 2026-02-03
refactor(ios): `resultForImage:`

GitToTheHub:
- Document meta data processing
- Remove any reference of `self.cdvUIImagePickerController` since it is not needed here and also not set, when `PHPickerViewController` is used
- Use `pictureOptions.sourceType` instead of `self.cdvUIImagePickerController.sourceType`
- Use `pictureOptions` parameter instead of `self.cdvUIImagePickerController.pictureOptions`
- Fix: Use local variable `imageDataToWrite` instead of `CDVCamera` property `self.data` to create the `CGImageSourceRef`. This code was wrong.

### Checklist

### DIFF
--- a/src/ios/CDVCamera.m
+++ b/src/ios/CDVCamera.m
@@ -740,19 +740,18 @@ static NSString* MIME_JPEG    = @"image/jpeg";
     return (scaledImage == nil ? image : scaledImage);
 }
 
-- (void)resultForImage:(CDVPictureOptions*)options
+- (void)resultForImage:(CDVPictureOptions*)pictureOptions
                   info:(NSDictionary*)info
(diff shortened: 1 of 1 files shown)


dpogue (commented):
Did we intentionally lose the `imageDataWithExif = [self.data mutableCopy];` path?

GitToTheHub:
I changed the code regarding the meta data processing and also fixed a bug where `self.data` was used to create the `CGImageSourceRef` for copying the image data from. Now the local variable `imageDataToWrite` is used.

To write the image it is not necessary to create a `mutableCopy` this was just to reuse the variable `imageDataWithExif` for the final write.

PR #969, opened 2026-07-25
chore(CDVCamera): refactor code when returning image after getting the location

GitToTheHub:
### Platforms affected

iOS

### Motivation and Context

- Renamed property `data` to `tempImageDataForLocationManager` since it is only used when getting location data for an image after capturing it
- Renamed method `imagePickerControllerReturnImageResult` to `returnImageAfterLocationUpdate` since it is only called, when a location was fetched, or an error ocurred while fetching one.

### Checklist


GitToTheHub:
If no one has any objections, I would merge this.

PR #873, opened 2023-12-06
feat: add stop API to close active camera or media picker

NityaSantosh26:
### Platforms affected
Android and iOS

### Motivation and Context

I have an usecase where user opens camera and if there's an high priority push notification is received, when user clicks, it doesn't close the camera, User has to manually close the camera. Now using the navigator.camera.stop(). I can achieve this behavior.
And there are other instances which needs closing of camera or photo library from the code.

### Description

Added a new method stop() which handles the closing of CAMERA or PHOTOLIBRARY for both Android and iOS.

### Testing

I have verified the usecase I have mentioned above, when a push notification is clicked that closes the camera and takes to the destination page in the module by calling the stop() method.

### Checklist

- [ ] I've run the tests to see all new and existing tests pass
- [ ] I added automated test coverage as appropriate for this change
- [ ] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [ ] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [ ] I've updated the documentation if necessary

PR #970, opened 2026-07-31
fix(android): clean up pending camera temp files on cancel/stop

GitToTheHub:
### Platforms affected

Android

### Motivation and Context

Follow up of PR #873.

@erisu noticed after PR #873 was merged, if there should be any cleanup procedures after `stop` was called. For example when `takePicture` is executed `this.imageUri` is set and gets not cleared. After checking this, it turned out, that this is the case, but this was also true, when an Activity was canceled by the user. This fixes both cases.

### Description

- delete pending image and cropped temp files when camera flow is stopped
- run cleanup when image selection/capture is canceled
- reset pending URI and file path state after best-effort cleanup

Generated-By: GPT-5.3-Codex

### Checklist

### DIFF
--- a/src/android/CameraLauncher.java
+++ b/src/android/CameraLauncher.java
@@ -327,6 +327,8 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
                 this.cordova.getActivity().finishActivity((sourceType + 1) * 16 + returnType + 1);
             }
         }
+
+        cleanupPendingResultFiles();
     }
 
     public void takePicture(int returnType, int encodingType)
@@ -901,6 +903,7 @@ public class CameraLauncher extends CordovaPlugin implements MediaScannerConnect
 
             }// If cancelled
             else if (resultCode == Activity.RESULT_CANCELED) {
+                cleanupPendingResultFiles();
                 this.failPicture("No Image Selected");
             }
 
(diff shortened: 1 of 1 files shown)


src/android/CameraLauncher.java
```diff
@@ -1214,6 +1219,37 @@ private void cleanup(Uri oldImage, Uri newImage, Bitmap bitmap) {
         System.gc();
     }
 
+    /**
+     * Best-effort cleanup for temporary files when capture/selection is cancelled or stopped.
+     */
+    private void cleanupPendingResultFiles() {
+        deleteIfExists(this.imageUri);
```
erisu, line 1226: I think the `imageUri` contains a `content://` URI, and the `deleteIfExists` method which uses `new File(filePath).delete()` would not work as it expects a filesystem path.

I believe we would need to use the `ContentResolver` to delete the file by passing in the `imageUri`.

src/android/CameraLauncher.java
```diff
@@ -1214,6 +1219,37 @@ private void cleanup(Uri oldImage, Uri newImage, Bitmap bitmap) {
...
+        deleteIfExists(this.croppedUri);
```
erisu: I dont believe this is a content uri so it might be correct but needs confirmation.

